PR #47, opened 2022-09-19
fix #28 for warning code E241

### LINKED ISSUE #28: Add quick fix for "multiple spaces after ‘,’" error
Ref: https://github.com/microsoft/vscode-flake8/issues/14

There are two things that needs to be done:

- [ ] Create quick fix for warning code `E241` that triggers the `editor.action.formatDocument` action
- [ ] Add unit test

SruthiSathya:


karthiknadig:
@SruthiSathya that file is still not formatted. You don't really need to create a separate PR. You can just create a commit with formatting fixes. just install `black` and then run `black src/test/python_tests/test_code_actions.py` and that should fix it.

Issue #104, opened 2023-01-24
How to use flake8 plugins


wardy3:
Hi

I can't see how to use various flake8 plugins

For example, I use:
- flake8-2020 
- flake8-aaa 
- flake8-encodings 
- flake8-newspaper-style
- wemake-python-styleguide

How can I configure the extension to use these plugins? Currently, I am installing flake8 itself and these plugins in my venv and it's working ok in vscode but it keeps suggesting I use this extension

thanks for the good work!

karthiknadig:
@wardy3 Can you try setting `"flake8.importStrategy": "fromEnvironment"` and it should work?

wardy3:
Yes, this seems to be working for me. That was simple :) Thanks!

PR #134, opened 2023-03-17
enable flake8 configuration from pyproject.toml

dciborow:
Adding a flake8 extension that enables configuration of flake8 via pyproject.toml. Also updating the readme to better describe the setup process for users.

Closes https://github.com/microsoft/vscode-flake8/issues/135


README.md
```diff
@@ -14,6 +14,34 @@ Once installed in Visual Studio Code, flake8 will be automatically executed when
 
 If you want to disable flake8, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension) per workspace in Visual Studio Code.
 
+## Configuration
+`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file. The extension will first check for the configuration file in the root directory, either the `pyproject.toml`, before checking the `.flake8` file. You can specific a custom location using the flake8.args as either `--toml-config=/path/to/pyproject.toml` or `--config=/path/to/.flake8`. While flake8 does not natively support the pyproject.toml, support is provided by the [flake8-pyproject](https://pypi.org/project/Flake8-pyproject/)) extension.
```
karthiknadig, line 18:
```suggestion
`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file. `flake8` will first check for the configuration file in the root directory, either the `pyproject.toml`, before checking the `.flake8` file. You can specifiy a custom location using the flake8.args as either `--toml-config=/path/to/pyproject.toml` or `--config=/path/to/.flake8`. While flake8 does not natively support the pyproject.toml, support is provided by the [flake8-pyproject](https://pypi.org/project/Flake8-pyproject/) plugin included in the extension.
```

karthiknadig (changes requested):

dciborow:
> removed request for [luabud](https://github.com/luabud) and [cwebster-99](https://github.com/cwebster-99)

Did not mean to do @luabud  and @cwebster-99  for review as well please :-)

README.md
```diff
@@ -14,6 +14,34 @@ Once installed in Visual Studio Code, flake8 will be automatically executed when
 
 If you want to disable flake8, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension) per workspace in Visual Studio Code.
 
+## Configuration
+`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file. `flake8` will first check for the configuration file in the root directory, either the `pyproject.toml`, before checking the `.flake8` file. You can specifiy a custom location using the flake8.args as either `--toml-config=/path/to/pyproject.toml` or `--config=/path/to/.flake8`. While flake8 does not natively support the pyproject.toml, support is provided by the [flake8-pyproject](https://pypi.org/project/Flake8-pyproject/) plugin included in the extension.
```
luabud, line 18: just to make sure I understand it right, does flake8 check for pyproject.toml before checking for the .flake8 file? If so, I'd only suggest changing the second sentence to something like "`flake8` will check for a configuration file in the root directory. The `pyproject.toml` file takes precedence over `.flake8`."  (or something like that).

README.md
```diff
@@ -14,6 +14,34 @@ Once installed in Visual Studio Code, flake8 will be automatically executed when
...
+
+The following is an example configuration for `flake8` in the `pyproject.toml`.
```
luabud, line 20:
```suggestion
The following is an example configuration for `flake8` in a `pyproject.toml` file.
```

README.md
luabud: Sounds good. Then maybe something like:
> "`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file located in your project's root directory"

Thoughts?

dciborow:
```suggestion
`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file. `flake8` will first check for the configuration file in the root directory, first checking the `pyproject.toml`, before checking the `.flake8` file. You can specify a custom location using the flake8.args as either `--toml-config=/path/to/pyproject.toml` or `--config=/path/to/.flake8`. While flake8 does not natively support the pyproject.toml, support is provided by the [flake8-pyproject](https://pypi.org/project/Flake8-pyproject/) plugin included in the extension.
```

README.md
```diff
@@ -14,6 +14,34 @@ Once installed in Visual Studio Code, flake8 will be automatically executed when
 
 If you want to disable flake8, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension) per workspace in Visual Studio Code.
 
+## Configuration
+`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file. `flake8` will first check for the configuration file in the root directory, first checking the `pyproject.toml`, before checking the `.flake8` file. You can specify a custom location using the flake8.args as either `--toml-config=/path/to/pyproject.toml` or `--config=/path/to/.flake8`. While flake8 does not natively support the pyproject.toml, support is provided by the [flake8-pyproject](https://pypi.org/project/Flake8-pyproject/) plugin included in the extension.
```
luabud, line 18:
```suggestion
`flake8` can be configured with either a `pyproject.toml` or [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html) file. `flake8` will check for the configuration file in the root directory, first checking the `pyproject.toml`, before checking the `.flake8` file. You can specify a custom location using the flake8.args as either `--toml-config=/path/to/pyproject.toml` or `--config=/path/to/.flake8`. While flake8 does not natively support the pyproject.toml, support is provided by the [flake8-pyproject](https://pypi.org/project/Flake8-pyproject/) plugin included in the extension.
```

luabud (approved):

brettcannon:
I don't know if I'm comfortable in taking this. This requires a 3rd-party library from flake8 to work, so we would have to track a second project for this to make sense. And the flake8 maintainers have made it very clear that they don't want to support `pyproject.toml`.

adzenith:
While I agree from a purity of flake8 standpoint this seems like not the best idea, I have to say I came here to the issues section because I wanted this exact functionality. I ended up just setting [`fromEnvironment`](https://github.com/microsoft/vscode-flake8/issues/104#issuecomment-1402968471) and installing this plugin myself, but it would be cool to have this work automatically so that a workspace can be shared with others and "just work" rather than requiring manual pip installs.
Perhaps there could instead just be a configurable list of plugins that the extension could install? That would be better from a purity of design standpoint, and also useful in many more scenarios, but would presumably be much harder to implement.

brettcannon:
> While I agree from a purity of flake8 standpoint this seems like not the best idea

Honestly the best idea would be for flake8 to add support for `pyproject.toml`, but the project's maintainers are pretty vehement that they don't want that. But we also have to balance this with the maintenance cost. If someone wants to fork this repo and create an extension that uses one of these helper libraries then please do so! Otherwise we want to keep this extension easy to maintain and that means keeping it somewhat simple.

I honestly would recommend looking at Ruff as a linter as it uses `pyproject.toml` and whose [VS Code extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) is maintained by the Ruff team themselves.

dciborow:
> > Honestly the best idea would be for flake8 to add support for `pyproject.toml`, but the project's maintainers are pretty vehement that they don't want that. But we also have to balance this with the maintenance cost.

I completely agree with you, but based on the repo maintainer gave up on that being possible.

I have also started using the Ruff linter. But find I still like to include my flake8 configuration in my pyproject.toml, particularly to disable overlapping warnings from flake8.
Some people may be using flake8 more widely than my project and may have the extension installed in vscode, and this lets me make sure they aren't spamming with problems in my projects.

Will regroup offline on how to move forward.

brettcannon:
> I have also started using the Ruff linter. But find I still like to include my flake8 configuration in my pyproject.toml, particularly to disable overlapping warnings from flake8.

If Ruff is a missing a lint you want from flake8, then I would ask Ruff to consider including it and getting off of flake8.



> Some people may be using flake8 more widely than my project and may have the extension installed in vscode, and this lets me make sure they aren't spamming with problems in my projects.

Unfortunately that's the potential cost of working outside of what flake8 wants by using `pyproject.toml` for your flake8 configuration; pick your poison.

I'm going to leave other comments on the issue, but I'm closing this (at least for now) as I don't think this is the direction we want to go.

Issue #141, opened 2023-04-10
flake8 error at column 0 causes lsp to break


PookieBuns:
![image](https://user-images.githubusercontent.com/83051474/230890430-bc09aa5a-1c77-48d0-a9bf-65eabe3e67fc.png)
When I run linting for this file, it raises a value error of Position.character out of range in the log.

After modifying the source code for the lsp_server, adding a check
![image](https://user-images.githubusercontent.com/83051474/230890843-938cb62c-844d-41df-9e27-49bd56bf5802.png)
I found out the error thrown is caused by a column value of zero in the following log
![image](https://user-images.githubusercontent.com/83051474/230890916-05c86dcf-6ff4-4b2b-8727-d959980509f1.png)

I'm not sure exactly why the _parse_output_using_regex function subtracts by one, but this causes column 0 errors to break the lsp

I would suggest adding a conditional statement to set column to 0 if it is less than 0?

vscode version = 1.77.1
windows version = windows 11 education 22621.1105
flake version = 6.0.0 (flake8-annotations: 3.0.0, mccabe: 0.7.0, pycodestyle: 2.10.0, pyflakes: 3.0.1) CPython 3.8.16 on Linux
flake8 v2023.4.0

karthiknadig:
Flake8 typically has column offset of 1, (i.e, its columns start at 1). In the above case the fix is to ensure that the value never gets negative.

Note that, the issue here comes from a pluggin `flake8-annotation` that seems to be using 0 based columns.

camilla122333:
can I work on this? ghc-osd

hhalaby:
Hello, any updates on this issue ?

I tried using the flake8 extension and this is the error output I get every time I restart the server:

```
2023-11-02 10:50:18.879 [info] Server run command: /home/halaby/.venvs/shake/bin/python /home/halaby/.vscode-server/extensions/ms-python.flake8-2023.8.0/bundled/tool/lsp_server.py
2023-11-02 10:50:18.879 [info] Server: Start requested.
2023-11-02 10:50:19.220 [info] Traceback (most recent call last):
  File "/home/halaby/.vscode-server/extensions/ms-python.flake8-2023.8.0/bundled/tool/lsp_server.py", line 251, in <module>
    def code_action(params: lsp.CodeActionParams) -> List[lsp.CodeAction]:
[...]

PR #143, opened 2023-04-10
Ensure column is pinned to 0

karthiknadig:
Fix for https://github.com/microsoft/vscode-flake8/issues/141

We have tests to validate standard flake8. We need this change to ensure some of the plugins don't fail. Authors of some plugins use 0 based column index when flake8 themselves use 1 based column index. From the LSP server we can;t differentiate which plugin generated the error, so we masure that the column value does not become negative.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,6 +14,9 @@
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
     "python.testing.cwd": "${workspaceFolder}",
-    "python.formatting.provider": "black",
-    "editor.formatOnSave": true
+    "python.formatting.provider": "none",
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -14,6 +14,9 @@
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
     "python.testing.cwd": "${workspaceFolder}",
-    "python.formatting.provider": "black",
-    "editor.formatOnSave": true
+    "python.formatting.provider": "none",
```
rzhao271, line 17: Is this change intentional?

karthiknadig: yes

Issue #195, opened 2023-08-02
Allow overriding severity based on error code prefix match


jnooree:
Pycodestyle error codes (which are reported by flake8) are grouped by prefix: `E1xx` for indentation, `E2xx` for whitespaces, etc.
Currently, this extension could override flake8 severity only if full error code matches. It would be great to specify such "prefix" in config, and override all matching error codes at once. This also matches the behavior of flake8 itself.

For example, the following config:

```json
{
    "flake8.severity": {
        "E": "Warning",
        "E1": "Information",
    }
}
```

Will report all `Exxx` errors as warnings, and all indentation related errors (`E1xx`) as information.

karthiknadig:
Duplicate of https://github.com/microsoft/vscode-flake8/issues/7

See https://github.com/microsoft/vscode-flake8/issues/7 for more details on this.

karthiknadig:
@jnooree The code for this can be found here:
https://github.com/microsoft/vscode-flake8/blob/6f638010c80d4ee17a4ad515c3666f6d7cea7796/bundled/tool/lsp_server.py#L125-L135

We are open to accept PRs on this :)

PR #196, opened 2023-08-02
Allow overriding severity based on error code prefix match

jnooree:
See #195 for context.

### DIFF
--- a/bundled/tool/lsp_server.py
+++ b/bundled/tool/lsp_server.py
@@ -133,7 +133,7 @@ def _get_severity(
         code = code[:-1]
 
     if value is None:
-        severity.get(code_type, "Error")
+        value = severity.get(code_type, "Error")
 
     try:
         return lsp.DiagnosticSeverity[value]
--- a/bundled/tool/lsp_server.py
+++ b/bundled/tool/lsp_server.py
@@ -126,7 +126,15 @@ def _get_severity(
     code: str, code_type: str, severity: Dict[str, str]
 ) -> lsp.DiagnosticSeverity:
     """Converts severity provided by linter to LSP specific value."""
-    value = severity.get(code, None) or severity.get(code_type, "Error")
+    value = None
+
+    while code and value is None:
+        value = severity.get(code, None)
+        code = code[:-1]
+
+    if value is None:
+        severity.get(code_type, "Error")
+
     try:
         return lsp.DiagnosticSeverity[value]
     except KeyError:
--- a/src/test/python_tests/test_linting.py
+++ b/src/test/python_tests/test_linting.py
@@ -195,7 +195,7 @@ def test_publish_diagnostics_on_close():
     assert_that(actual, is_(expected))
 
 
-@pytest.mark.parametrize("lint_code", ["F401", "F"])
+@pytest.mark.parametrize("lint_code", ["F401", "F40", "F4", "F"])
 def test_severity_setting(lint_code):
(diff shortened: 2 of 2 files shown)


karthiknadig:
@jnooree Thanks for the PR. Could you add tests for this scenario?

you can find an example here:
https://github.com/microsoft/vscode-flake8/blob/6f638010c80d4ee17a4ad515c3666f6d7cea7796/src/test/python_tests/test_linting.py#L198-L258

bundled/tool/lsp_server.py
```diff
@@ -126,7 +126,15 @@ def _get_severity(
...
-    value = severity.get(code, None) or severity.get(code_type, "Error")
+    value = None
+
+    while code and value is None:
+        value = severity.get(code, None)
+        code = code[:-1]
+
+    if value is None:
+        severity.get(code_type, "Error")
+
```
karthiknadig, line 137: This implementation could be redone as:
```python
def _get_severity(code: str, code_type: str, severity: Dict[str, str]) -> lsp.DiagnosticSeverity:
    """Converts severity provided by linter to LSP specific value."""

    try:
        while code:
            if code in severity:
                return lsp.DiagnosticSeverity[severity[code]]
            code = code[:-1]

        if code_type in severity:
            return lsp.DiagnosticSeverity[severity[code_type]]
    except KeyError:
        pass

    return lsp.DiagnosticSeverity.Error
```

bundled/tool/lsp_server.py
```diff
@@ -126,7 +126,15 @@ def _get_severity(
...
     """Converts severity provided by linter to LSP specific value."""
-    value = severity.get(code, None) or severity.get(code_type, "Error")
+    value = None
+
+    while code and value is None:
+        value = severity.get(code, None)
+        code = code[:-1]
+
+    if value is None:
+        severity.get(code_type, "Error")
```
karthiknadig, line 136:
```suggestion
        value = severity.get(code_type, "Error")
```

karthiknadig (changes requested):

bundled/tool/lsp_server.py
jnooree: This will require double lookup on the severity dictionary. If you don't mind that, I'll accept the changes.

karthiknadig: Yeah I agree. Restored to what you had with the fix for `value` with `code_type`: 
```python
def _get_severity(code: str, code_type: str, severity: Dict[str, str]) -> lsp.DiagnosticSeverity:
    """Converts severity provided by linter to LSP specific value."""
    value = None
    while code and value is None:
        value = severity.get(code, None)
        code = code[:-1]

    if value is None:
        value = severity.get(code_type, "Error")

    try:
        return lsp.DiagnosticSeverity[value]
    except KeyError:
        pass

    return lsp.DiagnosticSeverity.Error
```

jnooree: Great, thanks! The fix was applied in 8714960.

jnooree:
Do the tests added in 670dccb suffice?

karthiknadig:
> Do the tests added in [670dccb](https://github.com/microsoft/vscode-flake8/commit/670dccb69defe7884c83a001190597d5957ba6c3) suffice?

It looks good.

karthiknadig (approved):

anthonykim1 (approved):

PR #234, opened 2023-10-05
Update to latest version of packages and python

karthiknadig:


.github/workflows/push-check.yml
```diff
@@ -50,19 +50,19 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python: ['3.7', '3.8', '3.9', '3.10']
+        python: ['3.8', '3.8', '3.9', '3.10', '3.11', '3.12-dev']
```
eleanorjboyd, line 53: why are these removed?

PR #268, opened 2023-12-13
Add linting enable/disable functionality

karthiknadig:
For https://github.com/microsoft/vscode-python/issues/22607


bundled/tool/lsp_server.py
```diff
@@ -544,6 +550,11 @@ def _run_tool_on_document(
     # deep copy here to prevent accidentally updating global settings.
     settings = copy.deepcopy(_get_settings_by_document(document))
 
+    if not settings["enabled"]:
+        log_warning(f"Skipping file [Linting Disabled]: {document.path}")
```
anthonykim1, line 554: is there reason why we have two separate log_warning? Is it to distinguish between f string and `

anthonykim1 (approved):

bundled/tool/lsp_server.py
karthiknadig: Long lines are hard to read in the logs.

PR #466, opened 2026-04-07
Refactor: expandTilde dedup, jsonrpc env+ConnectionError, _get_settings_by_path

edvilme:
Trivial normalization changes for shared package extraction. No behavioral changes.

> Tracking: microsoft/vscode-flake8#465 | microsoft/vscode-python-tools-extension-template#290

- [x] **4.0f** Deduplicate `expandTilde` (settings.ts -> envFile.ts)
- [x] **4.2b** Add `env` parameter to `start_process()` in `lsp_jsonrpc.py`
- [x] **4.2c** `lsp_jsonrpc.py`: `Exception` -> `ConnectionError`
- [x] **4.S2** Add `_get_settings_by_path()` to `lsp_server.py`

### DIFF
--- a/bundled/tool/lsp_server.py
+++ b/bundled/tool/lsp_server.py
@@ -697,6 +697,23 @@ def _get_settings_by_document(document: TextDocument | None):
     }
 
 
+def _get_settings_by_path(file_path: pathlib.Path):
+    # Note: WORKSPACE_SETTINGS is keyed by normalized workspaceFS path
+    # (see _update_workspace_settings), so dict lookup by workspaceFS value is correct.
+    workspaces = {s["workspaceFS"] for s in WORKSPACE_SETTINGS.values()}
+
+    while file_path != file_path.parent:
+        str_file_path = utils.normalize_path(file_path)
+        if str_file_path in workspaces:
+            return WORKSPACE_SETTINGS[str_file_path]
+        file_path = file_path.parent
+
+    setting_values = list(WORKSPACE_SETTINGS.values())
+    if not setting_values:
+        return {}
+    return setting_values[0]
+
+
 # *****************************************************
 # Internal execution APIs.
 # *****************************************************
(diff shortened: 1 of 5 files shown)


rchiodo:
There's still a bunch of unresolved comments?

edvilme:
Both comments addressed:

1. **Key/value mismatch**: Added clarifying comment. WORKSPACE_SETTINGS is keyed by normalized workspaceFS path (see _update_workspace_settings lines 640-657), so the dict lookup is correct. The key IS the workspaceFS value.

2. **Unguarded [0] index**: Added empty guard: if not setting_values: return {}

bundled/tool/lsp_server.py
```diff
@@ -697,6 +697,23 @@ def _get_settings_by_document(document: TextDocument | None):
     }
```
rchiodo, line 698: Copilot generated:
-706

**Prior feedback unresolved — shadow-set pattern is structurally fragile (all four reviewers agree).** The comment asserting keys equal `workspaceFS` values doesn't prevent breakage if normalization ever diverges (trailing slash, case, slash direction on Windows). The Skeptic demonstrates a concrete scenario: if `normalize_path` produces backslashes but `workspaceFS` stores forward slashes, the `in workspaces` check silently fails on every ancestor and the function falls through to returning an arbitrary workspace's settings — a silent wrong-answer bug.

Replace the shadow-set + key-lookup with direct iteration:
```python
def _get_settings_by_path(file_path: pathlib.Path):
    while file_path != file_path.parent:
        str_file_path = utils.normalize_path(file_path)
        for _key, settings in WORKSPACE_SETTINGS.items():
            if settings["workspaceFS"] == str_file_path:
                return settings
        file_path = file_path.parent

    setting_values = list(WORKSPACE_SETTINGS.values())
    if not setting_values:
        return {}
    return setting_values[0]
```

rchiodo:
PR description claims "no behavioral changes" — Not addressed. The USE_DEBUGPY gate is a behavioral change for developers who had DEBUGPY_PATH set without USE_DEBUGPY. All four reviewers agree the description should acknowledge this.

edvilme:
Fixed in latest push: replaced shadow-set + key-lookup with direct iteration over WORKSPACE_SETTINGS.items(), exactly as suggested.

rchiodo (approved):
Approved via Review Center.

PR #477, opened 2026-04-10
Remove duplicate mock setup from test_get_cwd.py, rely on conftest

edvilme:
Add shared conftest.py cleanup: remove redundant _setup_mocks() from test_get_cwd.py that was poisoning test_change_cwd.py when both ran in the same session.

Changes:
- Removed 130-line duplicate _setup_mocks() from test_get_cwd.py
- Tests now rely on existing conftest.py for mock setup

All 92 tests pass.

Part of microsoft/vscode-python-tools-extension-template#290


StellaHuang95 (approved):

rchiodo (approved):
Approved via Review Center.

PR #478, opened 2026-04-10
test: add test_diagnostic_regex.py for output parsing

edvilme:
Add tests for the diagnostic regex pattern used to parse flake8 output into LSP diagnostics. Based on mypy's test_diagnostic_regex.py pattern.

Part of microsoft/vscode-python-tools-extension-template#290

### DIFF
--- a/src/test/python_tests/test_diagnostic_regex.py
+++ b/src/test/python_tests/test_diagnostic_regex.py
@@ -1,0 +1,186 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+"""
+Unit tests for diagnostic regex parsing in lsp_server.
+"""
+
+import pytest
+from lsp_server import DIAGNOSTIC_RE
+
+
+def _get_group_dict(line: str):
+    """Helper function to get match groups from a line."""
+    match = DIAGNOSTIC_RE.match(line)
+    if match:
+        return match.groupdict()
+    return None
+
+
+# ---------------------------------------------------------------------------
+# Positive-match tests (parametrized)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "line, expected",
+    [
+        pytest.param(
+            "14,5,E,E302:expected 2 blank lines, got 1",
+            {
+                "line": "14",
+                "column": "5",
+                "type": "E",
+                "code": "E302",
+                "message": "expected 2 blank lines, got 1",
+            },
+            id="E302",
+        ),
+        pytest.param(
+            "42,12,W,W503:line break before binary operator",
+            {
+                "line": "42",
+                "column": "12",
+                "type": "W",
+                "code": "W503",
+                "message": "line break before binary operator",
+            },
+            id="W503",
+        ),
+        pytest.param(
+            "1,1,F,F401:'os' imported but unused",
+            {
+                "line": "1",
+                "column": "1",
+                "type": "F",
+                "code": "F401",
+                "message": "'os' imported but unused",
+            },
+            id="F401",
+        ),
+        pytest.param(
+            "10,1,C,C901:'my_function' is too complex (15)",
+            {
+                "line": "10",
+                "column": "1",
+                "type": "C",
+                "code": "C901",
+                "message": "'my_function' is too complex (15)",
+            },
+            id="C901",
+        ),
+        pytest.param(
+            "7,80,E,E501:line too long (120 > 79 characters)",
+            {
+                "line": "7",
+                "column": "80",
+                "type": "E",
+                "code": "E501",
+                "message": "line too long (120 > 79 characters)",
+            },
+            id="E501",
+        ),
+        pytest.param(
+            # Known edge case: the regex captures "-1" as a string.  Downstream
+            # code that calls ``int(column)`` will get -1, which is an invalid
+            # LSP position; downstream should clamp to 0.
+            "5,-1,E,E303:too many blank lines (3)",
+            {
+                "line": "5",
+                "column": "-1",
+                "type": "E",
+                "code": "E303",
+                "message": "too many blank lines (3)",
+            },
+            id="negative-column",
+        ),
+        pytest.param(
+            "3,5,E,E111:indentation is not a multiple of four",
+            {
+                "line": "3",
+                "column": "5",
+                "type": "E",
+                "code": "E111",
+                "message": "indentation is not a multiple of four",
+            },
+            id="E111",
+        ),
+        pytest.param(
+            "15,20,W,W291:trailing whitespace",
+            {
+                "line": "15",
+                "column": "20",
+                "type": "W",
+                "code": "W291",
+                "message": "trailing whitespace",
+            },
+            id="W291",
+        ),
+        pytest.param(
+            "99999,1,E,E302:expected 2 blank lines, got 1",
+            {
+                "line": "99999",
+                "column": "1",
+                "type": "E",
+                "code": "E302",
+                "message": "expected 2 blank lines, got 1",
+            },
+            id="large-line-number",
+        ),
+        pytest.param(
+            "1,1,E,E999:SyntaxError: invalid syntax",
+            {
+                "line": "1",
+                "column": "1",
+                "type": "E",
+                "code": "E999",
+                "message": "SyntaxError: invalid syntax",
+            },
+            id="colon-in-message",
+        ),
+        pytest.param(
+            "0,0,E,E302:msg",
+            {"line": "0", "column": "0", "type": "E", "code": "E302", "message": "msg"},
+            id="zero-line-column",
+        ),
+        pytest.param(
+            "1,1,E,E302:",
+            {"line": "1", "column": "1", "type": "E", "code": "E302", "message": ""},
+            id="empty-message",
+        ),
+        pytest.param(
+            "1,1,e,E302:msg",
+            {"line": "1", "column": "1", "type": "e", "code": "E302", "message": "msg"},
+            id="lowercase-type",
+        ),
+    ],
+)
+def test_diagnostic_regex_positive(line, expected):
+    """Verify that valid flake8 output lines are parsed correctly."""
+    data = _get_group_dict(line)
+    assert data is not None, f"Regex should match: {line}"
+    assert data == expected
+
+
+# ---------------------------------------------------------------------------
+# Negative-match tests
+# ---------------------------------------------------------------------------
+
+
+def test_diagnostic_regex_no_match_on_empty_string():
+    """Test that an empty string does not match the regex."""
+    assert _get_group_dict("") is None
+
+
+def test_diagnostic_regex_no_match_on_garbage():
+    """Test that random text does not match the regex."""
+    assert _get_group_dict("this is not a diagnostic line") is None
+
+
+def test_diagnostic_regex_no_match_without_code():
+    """Test that a line without a valid code does not match."""
+    assert _get_group_dict("14,5,E,:missing code") is None
+
+
+def test_diagnostic_regex_no_match_leading_whitespace():
+    """Leading whitespace prevents match because .match() is start-anchored."""
+    assert _get_group_dict(" 14,5,E,E302:msg") is None


src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,199 @@
...
+# Licensed under the MIT License.
+"""
+Unit tests for diagnostic regex parsing in lsp_server.
+"""
+
+import pathlib
+import sys
+
+# Add the bundled tool directory to sys.path for importing lsp_server
+BUNDLED_TOOL_DIR = (
```
rchiodo: Copilot generated:
-15

**`sys.path.insert(0, ...)` should move to `conftest.py` with an idempotency guard.** Both the Skeptic and Architect flagged this independently. `insert(0, ...)` gives `bundled/tool/` highest import priority for the entire test process, risking module shadowing. This repo already uses `conftest.py` with `if str(path) not in sys.path` guards (see `build/automation/*/tests/conftest.py`). Follow that pattern:

```python
# conftest.py
import sys, pathlib
_BUNDLED = str(pathlib.Path(__file__).parent.parent.parent.parent / "bundled" / "tool")
if _BUNDLED not in sys.path:
    sys.path.insert(0, _BUNDLED)
```

Also: the 4-level `parent.parent.parent.parent` traversal is fragile — if the test file moves, imports break silently at runtime.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,199 @@
...
+def _get_group_dict(line: str):
+    """Helper function to get match groups from a line."""
+    match = DIAGNOSTIC_RE.match(line)
+    if match:
+        return match.groupdict()
+    return None
+
+
+def test_diagnostic_regex_with_e302():
+    """Test that E302 (expected blank lines) errors are correctly parsed."""
```
rchiodo: Copilot generated:
-199

**The 12 positive-match tests are textbook `@pytest.mark.parametrize` candidates.** Each follows the identical pattern: define `line`, call `_get_group_dict`, assert groups. Collapsing them into a single parametrized test would reduce ~150 lines of boilerplate to ~30 lines and make adding new cases trivial (one tuple vs 12 lines). Keep the 3 negative-match tests as separate functions since they test a different concern.

```python
@pytest.mark.parametrize("line, expected", [
    ("14,5,E,E302:expected 2 blank lines, got 1", {"line": "14", "column": "5", "type": "E", "code": "E302", "message": "expected 2 blank lines, got 1"}),
    # ... remaining cases
])
def test_diagnostic_regex_positive(line, expected):
    data = _get_group_dict(line)
    assert data is not None
    assert data == expected
```

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def test_diagnostic_regex_with_negative_column():
+    """Test that lines with negative column values are correctly parsed."""
+    line = "5,-1,E,E303:too many blank lines (3)"
+
+    data = _get_group_dict(line)
+
+    assert data is not None, "Regex should match line with negative column"
+    assert data["column"] == "-1"
```
rchiodo: Copilot generated:
-111

**Negative column test gives false confidence.** The Skeptic notes that while the regex correctly captures `"-1"` as a string, the downstream consumer will do `int("-1")` producing `character: -1`, which is invalid per the LSP spec (positions are 0-based unsigned). This test validates regex capture but implicitly suggests negative columns "work" end-to-end. At minimum, add a comment documenting that this is a known edge case that the regex intentionally passes through, and that downstream code should clamp to 0.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,199 @@
...
+    """Test that an empty string does not match the regex."""
+    data = _get_group_dict("")
+
+    assert data is None, "Regex should not match an empty string"
+
+
+def test_diagnostic_regex_no_match_on_garbage():
+    """Test that random text does not match the regex."""
+    data = _get_group_dict("this is not a diagnostic line")
+
```
rchiodo: Copilot generated:
-181

**Quoted-line test doesn't test what it claims.** The comment says `_parse_output_using_regex` handles quote-stripping, but this test manually strips the quotes (`line[1:-1]`) and then tests the inner content — which is exactly what every other positive test already proves. This test adds zero incremental coverage. Either test the actual `_parse_output_using_regex` function's quote-stripping behavior, or remove this test to avoid misleading coverage.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,199 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
rchiodo, line 1: Copilot generated:
📍 src/test/python_tests/test_diagnostic_regex.py (file-level)

**Missing edge case coverage.** The Skeptic identified several realistic edge cases with no coverage:
- `"0,0,E,E302:msg"` — zero line/column (0-indexed output)
- `"1,1,E,E302:"` — empty message after colon
- `" 14,5,E,E302:msg"` — leading whitespace (`.match()` is start-anchored, so this silently fails)
- `"1,1,e,E302:msg"` — lowercase type field

These matter because silent non-matches mean silently dropped diagnostics for the user.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,199 @@
...
+import pathlib
+import sys
+
+# Add the bundled tool directory to sys.path for importing lsp_server
+BUNDLED_TOOL_DIR = (
+    pathlib.Path(__file__).parent.parent.parent.parent / "bundled" / "tool"
+)
+sys.path.insert(0, str(BUNDLED_TOOL_DIR))
+
+from lsp_server import DIAGNOSTIC_RE  # noqa: E402
```
rchiodo: Copilot generated:
**Verify the import target exists in the target repo.** The Architect could not find `bundled/tool/lsp_server.py` with a `DIAGNOSTIC_RE` export in this repository. The PR description references `microsoft/vscode-python-tools-extension-template#290` — confirm this PR is targeting the correct repo and that `DIAGNOSTIC_RE` is defined where expected.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Fixed — removed duplicate sys.path.insert; conftest.py already handles bundled/tool path setup.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Fixed — collapsed 12 positive-match tests into a single `@pytest.mark.parametrize` test with 13 cases (including new edge cases).

src/test/python_tests/test_diagnostic_regex.py
edvilme: Fixed — added comment noting downstream `int('-1')` produces invalid LSP position and should be clamped to 0.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Fixed — removed the quoted-line test since it only tested already-covered unquoted content, not actual quote-stripping behavior.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Fixed — added 4 edge cases: zero line/column, empty message after colon, leading whitespace (negative match), and lowercase type field.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Confirmed — `DIAGNOSTIC_RE` is defined at `bundled/tool/lsp_server.py:329`. Import is correct.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,186 @@
...
+import pytest
+from lsp_server import DIAGNOSTIC_RE
+
+
+def _get_group_dict(line: str):
+    """Helper function to get match groups from a line."""
+    match = DIAGNOSTIC_RE.match(line)
+    if match:
+        return match.groupdict()
+    return None
```
rchiodo, line 16: Copilot generated:
**Critical: Import target does not exist in this repository.** All four reviewers confirmed that `bundled/tool/lsp_server.py` with a `DIAGNOSTIC_RE` export is not present in this repo. The PR references `microsoft/vscode-python-tools-extension-template#290` — either this PR targets the wrong repository, or the corresponding source file needs to be added first. As-is, every test will fail with `ModuleNotFoundError` at collection time.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+"""
+Unit tests for diagnostic regex parsing in lsp_server.
+"""
+
+import pytest
+from lsp_server import DIAGNOSTIC_RE
+
+
```
rchiodo, line 10: Copilot generated:
-14

**Unresolved: `sys.path.insert(0, ...)` should move to `conftest.py` with an idempotency guard.** This was flagged in the prior review and remains unchanged. The repo's established pattern (see `build/automation/*/tests/conftest.py`) uses `if str(path) not in sys.path` guards. Create `src/test/python_tests/conftest.py` following that pattern and remove the inline `sys.path` manipulation.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,186 @@
...
+# ---------------------------------------------------------------------------
+# Positive-match tests (parametrized)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "line, expected",
+    [
+        pytest.param(
```
rchiodo, line 27: Copilot generated:
-155

**Unresolved: Collapse 12 positive-match tests into `@pytest.mark.parametrize`.** Each follows the identical pattern (define line → call `_get_group_dict` → assert groups). This would reduce ~130 lines to ~30, make adding cases trivial, and — as the Rules reviewer noted — naturally fix the **partial assertion** problem in 4 of these tests that only check a subset of the 5 capture groups. Assert against the complete expected dict in the parametrized version.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,186 @@
...
+        ),
+        pytest.param(
+            "3,5,E,E111:indentation is not a multiple of four",
+            {
+                "line": "3",
+                "column": "5",
+                "type": "E",
+                "code": "E111",
+                "message": "indentation is not a multiple of four",
+            },
```
rchiodo, line 104: Copilot generated:
-111

**Partial assertions + missing LSP spec comment.** This test only checks `column`, `code`, and `message` — leaving `line` and `type` unchecked (a regression in either would pass silently). Additionally, the prior review noted that `int("-1")` produces an invalid LSP `character` value. Add a one-line comment acknowledging this is a known edge case that downstream code should clamp to 0, and assert all 5 groups.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,186 @@
...
+    ],
+)
+def test_diagnostic_regex_positive(line, expected):
+    """Verify that valid flake8 output lines are parsed correctly."""
+    data = _get_group_dict(line)
+    assert data is not None, f"Regex should match: {line}"
+    assert data == expected
+
+
+# ---------------------------------------------------------------------------
```
rchiodo, line 164: Copilot generated:
-181

**Unresolved: This test adds zero incremental coverage.** It manually strips quotes (`line[1:-1]`) then tests the inner string — identical to what every other positive test proves. The comment on L168 claims `_parse_output_using_regex` handles quote-stripping, but that function is never called here. Either rewrite this to test the actual `_parse_output_using_regex` quote-stripping behavior, or remove it entirely.

src/test/python_tests/test_diagnostic_regex.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
rchiodo, line 1: Copilot generated:
📍 src/test/python_tests/test_diagnostic_regex.py (file-level)

**Unresolved: Missing edge case coverage.** The prior review identified four realistic edge cases still untested:
- `"0,0,E,E302:msg"` — zero line/column
- `"1,1,E,E302:"` — empty message after colon
- `" 14,5,E,E302:msg"` — leading whitespace (`.match()` is start-anchored, silently fails → dropped diagnostics)
- `"1,1,e,E302:msg"` — lowercase type field

The leading-whitespace case is particularly important as it would silently drop user diagnostics.

src/test/python_tests/test_diagnostic_regex.py
edvilme: `DIAGNOSTIC_RE` is defined at `bundled/tool/lsp_server.py:329`: `DIAGNOSTIC_RE = re.compile(...)`. It's used at line 348. The import `from lsp_server import DIAGNOSTIC_RE` works correctly — all 17 tests pass in CI and locally.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Already resolved in the previous push — the inline `sys.path.insert` was removed. conftest.py handles bundled/tool path setup.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Already resolved — the 12 positive tests were collapsed into a single `@pytest.mark.parametrize` test with 13 cases. All assertions check the complete 5-group dict.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Already resolved — the negative-column test now asserts all 5 groups and includes a comment about downstream `int('-1')` producing an invalid LSP position.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Already resolved — the quoted-line test was removed as it added no incremental coverage.

src/test/python_tests/test_diagnostic_regex.py
edvilme: Already resolved — all 4 edge cases added: zero-line-column, empty-message, leading-whitespace (negative match), lowercase-type. See parametrized test cases.

rchiodo (approved):
Approved via Review Center.

PR #483, opened 2026-04-13
refactor: replace per-cell notebook linting with whole-notebook linting

edvilme:
Backport of microsoft/vscode-pylint#785 — replaces per-cell notebook linting with whole-notebook linting for cross-cell context.

## Changes

- **New file: \undled/tool/lsp_notebook.py\** — shared notebook helpers:
  - \NOTEBOOK_SYNC_OPTIONS\ (moved from lsp_server.py)
  - \uild_notebook_source()\ — concatenates all code cells into a single Python source
  - \
emap_diagnostics_to_cells()\ — maps combined-source diagnostics back to individual cells
  - \CellOffset\ / \CellMap\ — cell line tracking
  - \MAGIC_LINE_RE\ — replaces IPython magic lines with \pass\

- **\undled/tool/lsp_server.py\**:
  - Replace per-cell \_lint_notebook_cell()\ with \_linting_helper_notebook()\ that lints all cells as one source
  - Remove \is_notebook\ parameter from \_linting_helper()\
  - Import \lsp_notebook\ and \	ypes\ modules

- **New tests**:
  - \	est_notebook_magic_lines_no_syntax_error\ — magic lines (\%\, \!\) don't cause E999
  - \	est_notebook_cross_cell_import_not_flagged_unused\ — \import os\ in cell 0 not flagged F401 when used in cell 1

## Why

Per-cell linting misses cross-cell context (e.g., imports in one cell used in another are flagged as unused). Whole-notebook linting concatenates all cells before running flake8, so cross-cell references are resolved correctly.

All 8 notebook tests pass.


bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+
+from __future__ import annotations
+
+import dataclasses
+import re
+from typing import Callable, Optional, Sequence
+
+from lsprotocol import types as lsp
+
+# Matches IPython magic lines (%, %%, !, !!) so they can be replaced with `pass`.
```
heejaechang, line 36: Copilot generated:
**Bug: `MAGIC_LINE_RE` matches valid Python continuation lines, producing false-positive E999.** The regex `^\s*[%!]` matches any line starting with `%` or `!`, including legitimate continuation lines:
```python
message = ("Hello %s!"
           % name)  # ← matched, replaced with `pass` → syntax error
```
The Advocate argued these are "exceedingly rare," but the Skeptic provided concrete examples that are valid Python. This produces the exact class of false positive this PR aims to fix.

**Suggested fix:** Require `%`/`!` to be followed by a word character (magic command name): `r"^\s*(%{1,2}\w|!(?!=)\w)"` — this excludes `%` as modulo operator and `!=` comparisons.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+    diagnostics: Sequence[lsp.Diagnostic],
+    cell_map: CellMap,
+) -> dict[str, list[lsp.Diagnostic]]:
+    """Map combined-source diagnostics back to individual cell URIs.
+
+    Each diagnostic's line range is adjusted relative to the owning cell.
+    Diagnostics whose start line doesn't fall in any cell are discarded.
+    If a diagnostic's end line crosses a cell boundary it is clamped.
+    """
+    per_cell: dict[str, list[lsp.Diagnostic]] = {
```
heejaechang, line 143: Copilot generated:
-132

**Diagnostic fields silently dropped during remapping.** The new `lsp.Diagnostic` constructor copies `message`, `severity`, `code`, `code_description`, `source` — but drops `tags`, `related_information`, and `data`. If flake8 (or the output parser) ever sets `DiagnosticTag.Unnecessary` for unused imports, the tag is lost and the editor won't render strike-through styling. Both the Skeptic and Advocate flagged this.

**Suggested fix:** Copy all fields from the original diagnostic and only override `range`, e.g., via `dataclasses.replace()` if `Diagnostic` supports it, or by explicitly forwarding all known fields.

bundled/tool/lsp_server.py
```diff
@@ -237,21 +200,73 @@ def _is_supported_file(document: TextDocument) -> bool:
...
+
+        per_cell = notebook.remap_diagnostics_to_cells(combined_diagnostics, cell_map)
+
+        # Publish per-cell diagnostics; cells with no issues get an empty list
+        # so that stale diagnostics from a previous run are cleared.
+        for cell_uri, diags in per_cell.items():
+            LSP_SERVER.text_document_publish_diagnostics(
+                lsp.PublishDiagnosticsParams(uri=cell_uri, diagnostics=diags)
+            )
+    except Exception:  # pylint: disable=broad-except
```
heejaechang, line 276: Copilot generated:
-269

**Broad `except` swallows errors without clearing stale diagnostics.** When `_linting_helper_notebook` throws, diagnostics from the *previous* successful run persist on all cells — the user sees stale results with no indication that linting broke. While the Advocate correctly notes that broad-except is standard for LSP notification handlers (to prevent server crashes), the handler should publish empty diagnostics for all known cells in the `except` block before returning, so stale results are cleared.

bundled/tool/lsp_server.py
```diff
@@ -237,21 +200,73 @@ def _is_supported_file(document: TextDocument) -> bool:
...
+        if not cell_map:
+            for cell in nb.cells:
+                if cell.kind == lsp.NotebookCellKind.Code and cell.document:
+                    LSP_SERVER.text_document_publish_diagnostics(
+                        lsp.PublishDiagnosticsParams(uri=cell.document, diagnostics=[])
+                    )
+            return
+
+        # Build a synthetic document pointing at the notebook's .ipynb path so
+        # that settings resolution and flake8 invocation work correctly.
```
heejaechang, line 229: Copilot generated:
-229

**Fragile duck-type: `SimpleNamespace` as synthetic document.** All three non-Rules reviewers flagged this. `_run_tool_on_document` and `_get_settings_by_document` access document attributes implicitly. If either starts accessing `.lines`, `.filename`, or any other `TextDocument` attribute, this breaks with `AttributeError` — which is silently swallowed by the broad `except Exception` at line ~268, resulting in zero diagnostics with no user-visible indication of failure.

The inline NOTE comment acknowledges the risk but defers. The Architect recommends replacing with a `@dataclass` or `Protocol` now while the attribute contract is small (5 fields). This is a new-feature PR where integration-point robustness is high-weight.

bundled/tool/lsp_server.py
```diff
@@ -237,21 +200,73 @@ def _is_supported_file(document: TextDocument) -> bool:
...
+        # NOTE: SimpleNamespace is used here as a lightweight stand-in for
+        # TextDocument. If _run_tool_on_document or
+        # _get_settings_by_document begin accessing additional attributes,
+        # consider replacing this with a Protocol or TypedDict.
+        nb_path = _get_document_path(notebook_uri)
+        combined_doc = types.SimpleNamespace(
+            uri=notebook_uri,
+            path=nb_path,
+            source=combined_source,
+            language_id="python",
```
heejaechang: Copilot generated:
**`--stdin-display-name` receives an `.ipynb` path.** The Skeptic noted that `combined_doc.path` is set to the `.ipynb` file path. If `_run_tool_on_document` passes this as `--stdin-display-name`, flake8's `per-file-ignores`, `filename` globs, and `exclude` patterns may not match `.ipynb` as expected, potentially applying wrong configuration. Consider using a `.py` extension for the display name so flake8 treats it as a normal Python file.

bundled/tool/lsp_server.py
```diff
@@ -237,21 +200,73 @@ def _is_supported_file(document: TextDocument) -> bool:
     return False
```
heejaechang, line 201: Copilot generated:
-270

**Duplicated lint orchestration pipeline.** The Architect noted that `_linting_helper_notebook` reimplements the debounce → run → version-check → parse → publish pipeline from `_linting_helper`. If someone changes the parsing or debounce logic in one path, the other silently drifts. Consider extracting a shared `_run_lint_pipeline(document, key, ...)` that handles the common orchestration, with callers providing the document and publish strategy.

bundled/tool/lsp_server.py
```diff
@@ -262,10 +277,15 @@ def _clear_notebook_cell_diagnostics(cell_uri: str) -> None:
 
 
 def _linting_helper(
-    document: TextDocument, is_notebook: bool = False
+    document: TextDocument,
 ) -> list[lsp.Diagnostic]:
     try:
-        if not is_notebook and not _is_supported_file(document):
+        # Skip notebook cells — they are linted via _linting_helper_notebook
```
heejaechang, line 304: Copilot generated:
**Hardcoded URI scheme string.** Both the Skeptic and Architect flagged `str(document.uri).startswith("vscode-notebook-cell")` as a magic string. If VS Code changes the scheme or another editor host uses a different prefix, notebook cells silently fall through to standalone file linting (which will fail on unresolvable URIs).

**Suggested fix:** Extract to a named constant in `lsp_notebook.py` (e.g., `NOTEBOOK_CELL_URI_SCHEME = "vscode-notebook-cell"`) and reference it here.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+    source_parts: list[str] = []
+    cell_map: CellMap = []
+    current_line = 0
+
+    for cell in cells:
+        if cell.kind != lsp.NotebookCellKind.Code or cell.document is None:
+            continue
+        doc = get_text_document(cell.document)
+        if doc is None or doc.language_id != "python":
+            continue
```
rchiodo, line 99: Copilot generated:
**Bug — stale diagnostics persist on emptied cells.** When a cell's source is empty, `build_notebook_source` skips it (`if not source: continue`), so no `CellOffset` is created. Downstream, `remap_diagnostics_to_cells` only builds `per_cell` keys from `cell_map`, so this cell's URI never appears in the publish loop. Old diagnostics from before the cell was emptied remain pinned in the editor.

The same applies if a cell's `language_id` changes from Python to something else.

Fix: after the publish loop in `_linting_helper_notebook`, iterate all code cells and publish empty diagnostics for any cell URI not already in `per_cell`.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def remap_diagnostics_to_cells(
+    diagnostics: Sequence[lsp.Diagnostic],
+    cell_map: CellMap,
+) -> dict[str, list[lsp.Diagnostic]]:
+    """Map combined-source diagnostics back to individual cell URIs.
+
+    Each diagnostic's line range is adjusted relative to the owning cell.
+    Diagnostics whose start line doesn't fall in any cell are discarded.
```
rchiodo, line 140: Copilot generated:
**Diagnostic fields silently dropped during remap.** The new `lsp.Diagnostic(...)` constructor copies `range`, `message`, `severity`, `code`, `code_description`, `source` — but omits `tags`, `related_information`, and `data`. If flake8 plugins or future linter changes populate those fields, they are silently lost.

Either forward all fields (e.g., construct from the original and override only `range`), or explicitly document which fields are intentionally omitted and why.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+from __future__ import annotations
+
+import dataclasses
+import re
+from typing import Callable, Optional, Sequence
+
+from lsprotocol import types as lsp
+
+# Matches IPython magic lines (%, %%, !, !!) so they can be replaced with `pass`.
+MAGIC_LINE_RE = re.compile(r"^\s*[%!]")
```
rchiodo: Copilot generated:
**`MAGIC_LINE_RE` edge cases.** The Advocate notes `r"^\s*[%!]"` is a reasonable starting point but acknowledges edge cases: it could match `%` as a modulo operator at the start of a continuation line (false positive), and doesn't handle `?` (object introspection) or `??` (IPython help). The `pass` replacement is safe (worst case suppresses a real diagnostic, doesn't create false ones), but consider documenting these known limitations in a comment.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+    line_count: int
+
+
+CellMap = list[CellOffset]
+
+
+def build_notebook_source(
+    cells: list,
+    get_text_document: Callable[[str], Optional[object]],
+) -> tuple[str, CellMap]:
```
rchiodo, line 73: Copilot generated:
**Loose parameter typing.** `cells: list` should be `list[lsp.NotebookCell]`, and `get_text_document` return type `Optional[object]` should use a more precise type (or the proposed `LintableDocument` Protocol). This would give static analysis tools something to work with.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
...
+
+        sanitized_lines = [
+            "pass\n" if MAGIC_LINE_RE.match(line) else line for line in lines
+        ]
+
+        cell_map.append(CellOffset(cell.document, current_line, len(sanitized_lines)))
+        source_parts.extend(sanitized_lines)
+        current_line += len(sanitized_lines)
+
+    return "".join(source_parts), cell_map
```
rchiodo, line 118: Copilot generated:
**`get_cell_for_line` uses linear scan.** O(n) per diagnostic × O(diagnostics) total = O(n×d). Since `cell_map` is sorted by `start_line`, `bisect` would reduce this to O(d × log n). Not urgent for typical notebook sizes, but a quick win if performance becomes a concern.

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
rchiodo, line 1: Copilot generated:
📍 bundled/tool/lsp_notebook.py (module-level)

**No direct unit tests for `build_notebook_source` or `remap_diagnostics_to_cells`.** The integration tests cover the happy path, but edge cases in remapping (diagnostic spanning a cell boundary, diagnostic outside any cell, empty `cell_map`, cell with only magic lines) are untested. Unit tests with crafted inputs would strengthen confidence in the remap logic and would be fast to write since these are pure functions.

bundled/tool/lsp_server.py
```diff
@@ -237,21 +200,73 @@ def _is_supported_file(document: TextDocument) -> bool:
...
+
+        combined_diagnostics: Sequence[lsp.Diagnostic] = []
+        if result and result.stdout:
+            log_to_output(f"{notebook_uri} :\r\n{result.stdout}")
+            settings = copy.deepcopy(_get_settings_by_document(combined_doc))
+            combined_diagnostics = _parse_output_using_regex(
+                result.stdout, severity=settings["severity"]
+            )
+
+        per_cell = notebook.remap_diagnostics_to_cells(combined_diagnostics, cell_map)
```
rchiodo, line 256: Copilot generated:
📍 bundled/tool/lsp_server.py:_linting_helper_notebook (except block)

**Broad `except Exception` swallows errors without clearing stale diagnostics.** If any exception occurs (e.g., `AttributeError` from the `SimpleNamespace`, a subprocess crash), this logs the error but leaves diagnostics from the previous successful run pinned. The user sees stale errors with no indication linting is broken.

The Skeptic and Architect both flagged this independently. At minimum, publish empty diagnostics for all known cells inside the `except` block so stale results are cleared. Also consider narrowing the exception type.

bundled/tool/lsp_server.py
rchiodo, line 229: Copilot generated:
**`SimpleNamespace` synthetic document is fragile.** While the Advocate correctly notes the inline comment acknowledges this trade-off, both the Skeptic and Architect argue that a comment doesn't prevent the failure mode — if `_run_tool_on_document` or `_get_settings_by_document` accesses an attribute not on the namespace, it raises `AttributeError` at runtime, caught by the broad `except` and silently swallowed.

Define a minimal `Protocol` (e.g., `class LintableDocument(Protocol): uri: str; path: str; source: str; language_id: str; version: int`) so static analysis catches drift. This is a small change that eliminates a class of silent failures.

bundled/tool/lsp_server.py
```diff
@@ -177,48 +163,25 @@ def did_close(params: lsp.DidCloseTextDocumentParams) -> None:
...
-            _lint_notebook_cell(cell_document.uri)
-
-    if structure and structure.did_close:
-        for cell_document in structure.did_close:
-            _clear_notebook_cell_diagnostics(cell_document.uri)
+    """Re-lint all cells when any cell changes (for cross-cell context)."""
+    if params.change is not None and params.change.cells is not None:
+        structure = params.change.cells.structure
+        if structure and structure.did_close:
+            for cell_document in structure.did_close:
```
rchiodo, line 177: Copilot generated:
📍 bundled/tool/lsp_server.py:notebook_did_change

**No time-based debounce — potential subprocess storms during rapid typing.** The version check discards *stale results* but doesn't prevent *spawning redundant flake8 processes*. Every keystroke in any cell triggers `_linting_helper_notebook`, which spawns a flake8 subprocess over the entire notebook. For notebooks with many cells and rapid editing, this creates many concurrent processes where all but the last are wasted work.

Both the Skeptic and Architect flagged this independently. Consider adding a short idle delay (e.g., 200-300ms) before spawning, or cancelling in-flight processes when a newer version arrives.

bundled/tool/lsp_server.py
```diff
@@ -262,10 +277,15 @@ def _clear_notebook_cell_diagnostics(cell_uri: str) -> None:
 
 
 def _linting_helper(
-    document: TextDocument, is_notebook: bool = False
+    document: TextDocument,
 ) -> list[lsp.Diagnostic]:
     try:
```
rchiodo, line 303: Copilot generated:
📍 bundled/tool/lsp_server.py:_linting_helper (notebook cell guard)

**URI scheme check is brittle.** `str(document.uri).startswith("vscode-notebook-cell")` is a hard-coded string prefix. Other notebook providers (e.g., `untitled-notebook-cell`, Azure Data Studio) use different schemes and would bypass this guard, causing cells to be linted both individually and via `_linting_helper_notebook` — producing duplicate diagnostics. Consider a more robust check (e.g., querying the workspace for notebook document association).

bundled/tool/lsp_server.py
edvilme: Acknowledged. Tracking comment in code. Protocol type will come with the shared package extraction.

bundled/tool/lsp_server.py
edvilme: Intentional — flake8 uses the .ipynb path for config file discovery. Display name only affects error messages.

bundled/tool/lsp_server.py
edvilme: Will be consolidated in the shared package (vscode-common-python-lsp P3).

bundled/tool/lsp_notebook.py
edvilme: Notebooks typically have <50 cells. Linear scan is O(cells*diagnostics) — negligible.

bundled/tool/lsp_notebook.py
edvilme: Agreed — unit tests will be added in the shared package test suite.

bundled/tool/lsp_server.py
edvilme: Version-check debounce discards stale results, which is the right strategy here. Time-based debounce would delay feedback.

bundled/tool/lsp_notebook.py
edvilme: Improved regex to exclude != after % and require actual magic prefixes.

bundled/tool/lsp_notebook.py
edvilme: Fixed — added tags, related_information, and data fields.

bundled/tool/lsp_server.py
edvilme: Fixed — except block now clears stale diagnostics for all cells.

bundled/tool/lsp_server.py
edvilme: Fixed — extracted NOTEBOOK_CELL_SCHEME constant.

bundled/tool/lsp_notebook.py
edvilme: Already fixed — empty cells now get cleared diagnostics.

bundled/tool/lsp_notebook.py
edvilme: Already fixed — all diagnostic fields now forwarded.

bundled/tool/lsp_notebook.py
edvilme: Improved regex to be more precise about magic syntax.

bundled/tool/lsp_notebook.py
edvilme: Fixed — added Protocol type for type safety.

bundled/tool/lsp_server.py
edvilme: Tracking comment already in code. Will use Protocol in shared package.

bundled/tool/lsp_server.py
edvilme: Fixed — except block now clears stale diagnostics.

bundled/tool/lsp_server.py
edvilme: Fixed — extracted to NOTEBOOK_CELL_SCHEME constant.

bundled/tool/lsp_notebook.py
edvilme: Fixed — added 27 unit tests in bundled/tool/tests/test_lsp_notebook.py covering build_notebook_source (12 tests), get_cell_for_line (5 tests), and remap_diagnostics_to_cells (10 tests). Includes edge cases for magic lines, empty cells, cross-cell clamping, inverted range guard, and diagnostic field preservation.

bundled/tool/lsp_server.py
edvilme: Fixed — replaced SimpleNamespace with a typed SyntheticDocument dataclass in lsp_notebook.py (uri, path, source, language_id, version). lsp_server.py now uses notebook.SyntheticDocument(...) with defaults for language_id and version. Also added 3 tests for SyntheticDocument (30 total).

edvilme: Fixed — replaced SimpleNamespace with a typed SyntheticDocument dataclass in lsp_notebook.py (uri, path, source, language_id, version). lsp_server.py now uses notebook.SyntheticDocument(...) with defaults for language_id and version. Also added 3 tests for SyntheticDocument (30 total).

bundled/tool/lsp_server.py
```diff
@@ -237,21 +201,86 @@ def _is_supported_file(document: TextDocument) -> bool:
...
-    if document.language_id != "python":
-        return
-    diagnostics: list[lsp.Diagnostic] = _linting_helper(document, is_notebook=True)
-    LSP_SERVER.text_document_publish_diagnostics(
-        lsp.PublishDiagnosticsParams(uri=cell_uri, diagnostics=diagnostics)
-    )
+def _linting_helper_notebook(notebook_uri: str) -> None:
+    """Lint all code cells together and publish per-cell diagnostics."""
+    try:
+        nb = LSP_SERVER.workspace.get_notebook_document(notebook_uri=notebook_uri)
```
rchiodo, line 207: Copilot generated:
📍 bundled/tool/lsp_server.py:_linting_helper_notebook (empty-cells fast path)

**Bug: Race condition — `_lint_versions` not incremented before early return.** The `if not cell_map:` block publishes empty diagnostics and returns, but the version increment happens at L232 (after this return). An in-flight flake8 from a prior call (spawned when cells were non-empty) will pass its version check and re-publish stale diagnostics, overwriting the empty list.

All four reviewers independently confirmed this. Fix: move the version increment before the `build_notebook_source` call:

```python
with _lint_versions_lock:
    version = _lint_versions.get(notebook_uri, 0) + 1
    _lint_versions[notebook_uri] = version

combined_source, cell_map = notebook.build_notebook_source(...)
if not cell_map:
    # ... publish empty diagnostics and return
```

bundled/tool/lsp_notebook.py
```diff
@@ -0,0 +1,185 @@
...
+
+    uri: str
+    path: str
+    source: str
+    language_id: str = "python"
+    version: int = 0
+
+
+# Matches IPython magic lines (%, %%, !, !!) so they can be replaced with `pass`.
+MAGIC_LINE_RE = re.compile(r"^\s*(?:%%|%(?!=)|!)")
```
rchiodo: Copilot generated:
**Bug: `MAGIC_LINE_RE` matches valid Python continuation lines, producing false-positive E999.** Two vectors remain:

1. `%` as modulo: `"    % name)\n"` — `%(?!=)` passes because next char is space, not `=`. Line replaced with `pass` → SyntaxError.
2. `!` in `!=`: `"    != 0):\n"` — the `!` branch has no lookahead. Line replaced with `pass` → SyntaxError.

Both produce the exact class of false positive this PR exists to eliminate. Requiring a word character after `%`/`!` anchors on magic command names while excluding operators:

```python
MAGIC_LINE_RE = re.compile(r"^\s*(?:%%\w|%(?!=)\w|!(?!=)\w)")
```

bundled/tool/tests/test_lsp_notebook.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def test_cell_with_none_document(self):
+        cell = FakeCell("cell:0")
+        cell.document = None
+        source, cell_map = lsp_notebook.build_notebook_source(
+            [cell], lambda _: None
+        )
+        assert source == ""
+        assert cell_map == []
+
```
rchiodo, line 253: Copilot generated:
-232

**Test is a tautology — doesn't exercise the actual false-positive.** `"x %= 10\n"` starts with `x`, which fails the `^\s*` anchor, so the regex never matches regardless of the `%=` exclusion. This test passes trivially and gives false confidence.

Replace with tests for the actual vulnerable patterns:
```python
def test_magic_regex_does_not_match_modulo_continuation(self):
    cells = [FakeCell("cell:0")]
    docs = _make_docs({"cell:0": "    % name)\n"})
    source, _ = lsp_notebook.build_notebook_source(cells, docs)
    assert "% name)" in source  # must NOT be replaced with pass

def test_magic_regex_does_not_match_not_equal(self):
    cells = [FakeCell("cell:0")]
    docs = _make_docs({"cell:0": "    != 0)\n"})
    source, _ = lsp_notebook.build_notebook_source(cells, docs)
    assert "!= 0)" in source  # must NOT be replaced with pass
```

bundled/tool/lsp_notebook.py
edvilme: Fixed — added TextDocumentLike Protocol for the callback return type. The cells parameter can't use list[lsp.NotebookCell] without creating a direct pygls dependency, so it stays as list with a comment.

bundled/tool/lsp_server.py
edvilme: Fixed — except block now clears stale diagnostics for all notebook cells on error.

bundled/tool/lsp_server.py
edvilme: Fixed — replaced SimpleNamespace with a typed SyntheticDocument dataclass in lsp_notebook.py. lsp_server.py now uses notebook.SyntheticDocument(...).

bundled/tool/lsp_server.py
edvilme: Fixed — extracted NOTEBOOK_CELL_SCHEME constant. All URI scheme checks now reference it.

bundled/tool/lsp_notebook.py
edvilme: Fixed — added TextDocumentLike Protocol for the callback return type. cells parameter stays as list with a comment (can't import pygls type without dependency).

bundled/tool/lsp_server.py
edvilme: Fixed — except block now clears stale diagnostics for all notebook cells on error.

bundled/tool/lsp_server.py
edvilme: Fixed — replaced SimpleNamespace with typed SyntheticDocument dataclass in lsp_notebook.py.

bundled/tool/lsp_server.py
edvilme: Fixed — extracted NOTEBOOK_CELL_SCHEME constant. All URI scheme checks now reference it.

bundled/tool/lsp_server.py
edvilme: Fixed — moved version increment before build_notebook_source so the early return path also invalidates in-flight runs. Applied to all repos.

bundled/tool/lsp_notebook.py
edvilme: Fixed — tightened regex to require a word character after %/!: r'^\s*(?:%%\w|%(?!=)\w|!!|!(?!=)\w)'. This excludes % as modulo continuation, != as operator, and lone %/! with spaces. Added !! for shell capture.

bundled/tool/tests/test_lsp_notebook.py
edvilme: Fixed — replaced the tautological test with 4 proper tests: modulo continuation ('    %% name)'), != operator ('    != 0)'), %= assignment ('  %%= 10'), and !! shell capture. All exercise the actual vulnerable patterns.

rchiodo (dismissed):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #487, opened 2026-04-20
feat: integrate @vscode/common-python-lsp shared package

edvilme:
## Integrate `@vscode/common-python-lsp` shared package

Replaces duplicated infrastructure code with the shared `@vscode/common-python-lsp` package (npm + PyPI), eliminating the git submodule.

### What changed

**TypeScript (`src/common/*.ts`)** — 11 files replaced with thin wrappers:
- **Tier 1** (pure re-exports): `logging.ts`, `status.ts`
- **Tier 2** (re-exports + local config): `setup.ts`, `configWatcher.ts`
- **Tier 3** (re-exports + local functions for testability): `vscodeapi.ts`, `utilities.ts`, `envFile.ts`
- **Tier 4** (stateful wrappers with behavioral adaptation): `constants.ts`, `python.ts`, `settings.ts`, `server.ts`

**Python (`bundled/tool/`)** — 4 files replaced with thin wrappers:
- `lsp_utils.py`, `lsp_jsonrpc.py`, `lsp_notebook.py`, `lsp_runner.py`
- All delegate to `vscode_common_python_lsp` shared package

**Package source: npm + PyPI (no submodule)**
- npm: `@vscode/common-python-lsp@0.2.1`
- pip: `vscode-common-python-lsp==0.2.1`
- Removed `.gitmodules`, `submodules/` directory, `scripts/postinstall.js`
- `noxfile.py` installs from PyPI; pinned in `_update_npm_packages`

**Dependency cleanup:**
- Removed `dotenv`, `semver`, `@vscode/python-environments`, `@types/semver` (now transitive via shared package)
- Moved `fs-extra` to devDependencies (test-only)

**Test refactoring:**
- TS tests updated to stub at `vscode` module level (not intermediate wrappers) where needed for CommonJS compatibility
- Python tests updated for shared package imports; `MAX_LSP_CHARACTER` constant used in notebook tests
- Added `bundled/libs` to test `sys.path` for shared package resolution
- Uses public `reset_caches()` API from shared package for test isolation

**CI/Workflow cleanup:**
- Removed `submodules: true` from all checkout steps
- Removed empty `with:` blocks left from cleanup
- Cleaned `tsconfig.json` (`skipLibCheck` for `@sinonjs/fake-timers` DOM types, removed submodule exclude), `.vscodeignore`
- Regenerated `package-lock.json` (removed stale submodule entry)

### Design notes

- **Tier 3 local functions**: `isVirtualWorkspace` and `getDocumentSelector` are kept local because sinon cannot stub getter-backed CommonJS re-exports. Tests that need to stub these functions rely on them being regular function properties.
- **Tier 4 behavioral preservation**: Wrappers in `settings.ts` post-process `ignorePatterns` with `resolveVariables` and conditionally include `interpreter` — behavioral differences from the shared package defaults. `python.ts` uses a relay EventEmitter pattern that survives provider resets. `server.ts` adapts the shared options-bag API to the original positional-args signatures.
- **Zero behavioral changes**: All wrapper functions preserve the original signatures and behavior. The VSIX output changes only in that shared code is now resolved from `node_modules` / `bundled/libs` instead of inline.


bundled/tool/tests/test_lsp_notebook.py
```diff
@@ -335,7 +335,7 @@ def test_diag_spanning_cells_is_clamped(self, cell_map):
         d = result["cell:0"][0]
         assert d.range.start.line == 1
         assert d.range.end.line == 2  # clamped to max_end_line (3 - 1 = 2)
-        assert d.range.end.character == 0  # clamped → character=0
+        assert d.range.end.character == 2147483647  # clamped → end-of-line (LSP max)
```
edvilme: Why is this changed?

package.json
```diff
@@ -54,6 +54,7 @@
     "main": "./dist/extension.js",
     "l10n": "./l10n",
     "scripts": {
+        "postinstall": "tsc -p submodules/vscode-common-python-lsp/typescript/tsconfig.json",
```
rchiodo: Copilot generated:
✅ Verified: `tsc -p` will error if the submodule directory is missing, and there is no guard or fallback message. The `postinstall` script `tsc -p submodules/vscode-common-python-lsp/typescript/tsconfig.json` will fail if the submodule is not initialized (e.g., `git clone` without `--recurse-submodules`). This breaks `npm install` for every new contributor with no clear error message.

Additionally, this couples `npm install` to the submodule's TypeScript compilation — if the submodule's `tsconfig.json` changes or its `typescript` version differs, installs break unexpectedly.

**Suggested fix:** Either guard the script (`test -d submodules/vscode-common-python-lsp && tsc -p ... || echo 'Submodule not initialized; run git submodule update --init'`) or move the `tsc` step out of `postinstall` into a dedicated build step.

noxfile.py
```diff
@@ -24,6 +24,14 @@ def _install_bundle(session: nox.Session) -> None:
         "-r",
         "./requirements.txt",
     )
+    session.install(
```
rchiodo, line 27: Copilot generated:
-34

Unresolved from prior review: this installs the shared Python package into `bundled/libs/`, which ships inside the VSIX. The PR description still claims "zero behavioral changes," but this changes the VSIX artifact. Now that `src/common/` thin wrappers also exist, this install is logically consistent with the actual code — but the PR description needs to be updated to acknowledge both the VSIX change and the Stage 2 scope.

.github/workflows/pr-check.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
```
rchiodo, line 20: Copilot generated:
-21

Unresolved from prior review: `submodules: recursive` should be `submodules: true`. The shared submodule has no nested submodules, so `true` is sufficient and avoids accidentally pulling transitive submodules if the shared package ever adds one. This applies to all seven checkout steps across `pr-check.yml`, `push-check.yml`, and `codeql-analysis.yml`.

bundled/tool/tests/test_lsp_notebook.py
edvilme: The shared package uses `MAX_LSP_CHARACTER` (2^31-1, the LSP `uinteger` max) instead of `0` when clamping cross-cell diagnostic end positions. This means "end of line" per the LSP spec, which avoids zero-width or inverted ranges. The original `character=0` was technically incorrect — it placed the end of the range at the start of the line.

The test now asserts against the shared package's constant and includes a comment explaining the rationale.

package.json
edvilme: Fixed in b364f98. Moved the bare `tsc` call into `scripts/postinstall.js` which:
- Checks if the submodule tsconfig exists before compiling
- Prints a helpful warning if the submodule is not initialized
- Uses `npx tsc` so it works without a global TypeScript install

noxfile.py
edvilme: Updated the PR description in b364f98 to reflect the full Stage 1+2 scope, including the VSIX impact (shared Python package in `bundled/libs/`) and all the thin wrapper changes.

.github/workflows/pr-check.yml
edvilme: Fixed in b364f98. Changed all 9 occurrences across `pr-check.yml`, `push-check.yml`, and `codeql-analysis.yml` from `submodules: recursive` to `submodules: true`.

package.json
```diff
@@ -231,6 +231,7 @@
         "dotenv": "^17.4.1",
         "fs-extra": "^11.3.4",
         "semver": "^7.7.4",
+        "@vscode/common-python-lsp": "0.1.1",
```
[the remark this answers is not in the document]
edvilme: This comment references an older revision. The PR has since migrated from the git submodule to the published registry packages. The submodule, postinstall script, and all submodules checkout steps have been removed. The PR description already reflects this.

noxfile.py
```diff
@@ -24,6 +24,14 @@ def _install_bundle(session: nox.Session) -> None:
...
+        "-t",
+        "./bundled/libs",
+        "--no-cache-dir",
+        "--no-deps",
+        "--upgrade",
+        "vscode-common-python-lsp==0.1.1",
+    )
```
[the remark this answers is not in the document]
edvilme: Same as above — the submodule was removed in a later commit. The published PyPI package (vscode-common-python-lsp==0.1.1) is now the source of truth, matching the npm package version.

src/test/python_tests/conftest.py
```diff
@@ -140,6 +140,12 @@ def __init__(self, *args, **kwargs):
         sys.path.insert(0, tool_dir)
         _INJECTED_PATH = tool_dir
 
+    # Also add bundled/libs so the shared vscode_common_python_lsp package
+    # is importable (installed there by nox install_bundled_libs).
+    libs_dir = str(pathlib.Path(__file__).parents[3] / "bundled" / "libs")
+    if libs_dir not in sys.path:
+        sys.path.insert(0, libs_dir)
+
```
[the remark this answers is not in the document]
edvilme: Fixed in 5b17039. Changed from a single _INJECTED_PATH variable to an _INJECTED_PATHS list that tracks both tool_dir and libs_dir. Teardown now removes all injected paths in reverse insertion order.

tsconfig.json
```diff
@@ -5,6 +5,7 @@
         "lib": ["ES2020"],
         "sourceMap": true,
         "rootDir": "src",
+        "skipLibCheck": true,
```
rchiodo, line 12: Copilot generated:
⚠️ Unverified (files not in workspace; design claim assessed from diff only). `skipLibCheck: true` disables type-checking for **all** `.d.ts` files, including `@vscode/common-python-lsp`'s type declarations. This silences an entire class of type errors — if the shared package ships incorrect types (mismatched overloads, missing optional markers), the compiler won't catch it.

The Advocate argues this is standard npm practice, but the Architect notes that needing it is a red flag about the shared package's type quality. Since the entire PR strategy is "type-safe thin wrappers," disabling type-checking on the delegated-to package undermines the safety guarantee.

**Suggested fix:** Remove `skipLibCheck` and fix any type issues in the shared package. If that's infeasible short-term, add a `// TODO` comment explaining why it's needed and tracking removal.

bundled/tool/lsp_runner.py
```diff
@@ -32,39 +31,12 @@ def update_sys_path(path_to_add: str, strategy: str) -> None:
...
-
-        response = {"id": msg["id"], "error": result.stderr}
-        if is_exception:
-            response["exception"] = is_exception
-        elif result.stdout:
-            response["result"] = result.stdout
+from vscode_common_python_lsp import (  # noqa: E402
+    JsonRpc,
+    RunResult,
+    run_message_loop,
```
rchiodo, line 37: Copilot generated:
⚠️ Unverified (shared package `run_message_loop` not available in workspace for inspection). The old code explicitly prepended `""` to `sys.path` before running tool modules:
```python
with utils.substitute_attr(sys, "path", [""] + sys.path[:]):
    result = utils.run_module(...)
```
The new code calls `run_message_loop(RPC, run_module, RunResult)` without this manipulation. If the shared package's `run_message_loop` doesn't replicate the `[""] + sys.path[:]` substitution, any tool module that imports from CWD will fail with `ModuleNotFoundError`.

**Experiment needed:** Run a tool module that imports a sibling file from CWD and confirm it still resolves. If the shared package handles this internally, add a comment noting that.

src/test/python_tests/test_change_cwd.py
```diff
@@ -34,7 +34,10 @@ def test_change_cwd_permission_error_does_not_crash(caplog):
     original_cwd = os.getcwd()
     body_executed = False
 
-    with patch("lsp_utils.os.chdir", side_effect=PermissionError("Access denied")):
+    with patch(
```
rchiodo, line 40: Copilot generated:
✅ Verified (from diff evidence — patch targets visibly changed). Tests now patch `vscode_common_python_lsp.context.os.chdir` — an **internal module path** of the shared package. The diff shows the old stable target `lsp_utils.os.chdir` replaced with a cross-package internal path. If v0.2.0 reorganizes `context.py` → `runtime.py`, this test breaks.

Consider testing the *behavior* (verify CWD after call) rather than mocking the *mechanism*, or request that the shared package export test seams.

src/test/python_tests/test_stdlib_detection.py
```diff
@@ -183,22 +183,26 @@ def test_false_positive_site_packages_in_name():
 
 def test_user_site_packages_none_handling():
     """Test that None user site-packages is handled gracefully."""
-    # This test validates the fix for the case where site.getusersitepackages() returns None
-    # We can't easily simulate this, but we can verify the function handles None properly
+    # The shared package uses lru_cache; mock the underlying site function
```
rchiodo: Copilot generated:
✅ Verified (from diff evidence — internal coupling visible in diff). This test reaches into `vscode_common_python_lsp.paths._get_user_site_root.cache_clear()` — a private, underscored internal of the shared package. The diff also patches `_paths.site.getusersitepackages` directly.

**Suggested fix:** Add a defensive comment noting this coupling, or request the shared package expose a `reset_caches()` test helper.

tsconfig.json
edvilme: The only error without skipLibCheck is \@sinonjs/fake-timers\ (transitive via sinon) referencing the DOM \Performance\ type, which is unavailable in our Node/ES2020 lib target. Our own source code and the shared package types are still fully type-checked — skipLibCheck only affects third-party \.d.ts\ files. This is the same workaround vscode-isort uses. Added a comment explaining the rationale in 016a1af.

bundled/tool/lsp_runner.py
edvilme: Verified: the shared \
un_message_loop\ already handles this — line 65 of \process_runner.py\ wraps each \
un_module\ call with \substitute_attr(sys, 'path', [''] + sys.path[:])\. Added a comment in 016a1af documenting this.

src/test/python_tests/test_change_cwd.py
edvilme: The mock is necessary here — we're testing error-handling behavior when os.chdir raises PermissionError/OSError, which can't be reliably triggered with real paths cross-platform. Added comments in 016a1af documenting the mock target and its stability (change_cwd in context.py calls os.chdir directly, no intermediate wrapper).

src/test/python_tests/test_stdlib_detection.py
edvilme: Agreed. Created microsoft/vscode-common-python-lsp#19 which adds a public \
eset_caches()\ helper. Updated the test in 016a1af to use \rom vscode_common_python_lsp import reset_caches\ instead of reaching into private \_get_user_site_root.cache_clear()\.

src/common/server.ts
```diff
@@ -137,38 +36,19 @@ export async function restartServer(
         }
     });
     _disposables = [];
```
rchiodo, line 38: Copilot generated:
-48

**[unverified]** The old code had an explicit `try { await newLSClient.start() } catch { updateStatus(l10n.t('Server failed to start.'), Error) }` block that showed user-visible error feedback in the language status bar. The new delegation to `_restartServer` has no local try/catch — if it throws (misconfigured interpreter, missing server script, invalid CWD), `_disposables` is cleared but never repopulated, and no status-bar error is shown.

Either wrap the `_restartServer` call in try/catch with `updateStatus(l10n.t('Server failed to start.'), LanguageStatusSeverity.Error)`, or verify and document that the shared package handles this error path internally (including calling `updateStatus`).

[verified]

bundled/tool/lsp_notebook.py
```diff
@@ -1,190 +1,22 @@
...
+
+from vscode_common_python_lsp import (  # noqa: F401
+    MAGIC_LINE_RE,
+    NOTEBOOK_SYNC_OPTIONS,
+    CellOffset,
+    SyntheticDocument,
+    build_notebook_source,
+    get_cell_for_line,
+    remap_diagnostics_to_cells,
 )
```
rchiodo, line 18: Copilot generated:
This is production code importing from an internal submodule path: `from vscode_common_python_lsp.notebook import TextDocumentLike`. All other re-exports in this file use the top-level `vscode_common_python_lsp` namespace. If the shared package reorganizes its internal modules, this breaks at runtime. Request the shared package export `TextDocumentLike` from its top-level `__init__.py` (similar to how `reset_caches` was added upstream), or add a tracking comment/issue.

[verified]

src/common/python.ts
```diff
@@ -1,228 +1,58 @@
...
+    } catch {
+        return undefined;
     }
-    traceError(`Python version ${version?.major}.${version?.minor} is not supported.`);
-    traceError(`Selected python path: ${resolved?.executable.uri?.fsPath}`);
-    traceError(`Supported versions are ${PYTHON_VERSION} and above.`);
-    return false;
 }
 
 export function resetCachedApis(): void {
```
rchiodo, line 60: Copilot generated:
-57

**[unverified]** `resetCachedApis()` disposes the old `_provider` and creates a new one, but `initializePython` registered `onDidChangeEnvironment` listeners on the *old* provider's disposables list. After reset, the new provider may have no environment-change listener until `initializePython` is called again. If the extension lifecycle doesn't re-call `initializePython` after reset, interpreter changes stop triggering server restarts.

Verify that `resetCachedApis()` callers also re-invoke `initializePython`, or move the subscription into the `PythonEnvironmentsProvider` constructor so it's always active on new instances.

[verified]

src/common/server.ts
edvilme: The shared restartServer already handles this internally — it wraps createLanguageClient() and newLSClient.start() in try/catch blocks that call updateStatus(l10n.t('Server failed to start.'), LanguageStatusSeverity.Error) (server.ts lines 265-266, 276-277 in the shared package). No additional error handling needed in the wrapper.

bundled/tool/lsp_notebook.py
edvilme: Good catch. Added TextDocumentLike to the shared package top-level __init__.py exports and updated this import to use the public namespace. Fixed in 9809feb.

src/common/python.ts
edvilme: resetCachedApis() is only called from tests (not from extension.ts). The relay emitter pattern ensures interpreter-change events survive resets. Added a doc comment in 9809feb noting that callers should re-invoke initializePython() if environment-change listeners are needed on the new provider instance.

src/test/python_tests/test_stdlib_detection.py
```diff
@@ -183,22 +183,25 @@ def test_false_positive_site_packages_in_name():
 
 def test_user_site_packages_none_handling():
     """Test that None user site-packages is handled gracefully."""
-    # This test validates the fix for the case where site.getusersitepackages() returns None
-    # We can't easily simulate this, but we can verify the function handles None properly
+    from unittest.mock import patch
+
     import lsp_utils
+    from vscode_common_python_lsp import paths as _paths
+    from vscode_common_python_lsp import reset_caches
```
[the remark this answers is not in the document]
edvilme: Good catch — added bundled/libs to sys.path alongside bundled/tool so the standalone __main__ runner can resolve vscode_common_python_lsp. Fixed in 14ba2ef.

package.json
```diff
@@ -226,20 +226,17 @@
         ]
     },
     "dependencies": {
-        "@vscode/python-environments": "https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/npm/registry/@vscode/python-environments/-/python-environments-1.0.0.tgz",
         "@vscode/python-extension": "^1.0.6",
-        "dotenv": "^17.4.1",
-        "fs-extra": "^11.3.4",
-        "semver": "^7.7.4",
+        "@vscode/common-python-lsp": "0.2.1",
         "vscode-languageclient": "^8.1.0"
```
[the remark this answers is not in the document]
edvilme: Updated the PR description to reflect 0.2.1 throughout, and added Tier 4 details + dependency cleanup section. Fixed in 14ba2ef.

src/common/settings.ts
```diff
@@ -1,204 +1,75 @@
...
-    if (workspaceSetting.cwd.startsWith('~')) {
-        workspaceSetting.cwd = expandTilde(workspaceSetting.cwd);
+    const resolveInterpreter = includeInterpreter ? getInterpreterDetails : undefined;
+    const settings = (await _getWorkspaceSettings(
+        namespace,
+        workspace,
+        FLAKE8_TOOL_CONFIG,
+        resolveInterpreter,
+    )) as ISettings;
+
```
rchiodo, line 41: Copilot generated:
-44

[unverified] `ignorePatterns` is post-processed with `resolveVariables`, but `extraPaths` is not. The old code called `resolveVariables(getExtraPaths(namespace, workspace), workspace)` and then `expandTilde` on each entry. If the shared `_getWorkspaceSettings` doesn't handle variable resolution and tilde expansion for `extraPaths`, users with `${workspaceFolder}/lib` or `~/lib` in `flake8.extraPaths` will get broken `PYTHONPATH` values.

Suggested fix: add equivalent post-processing for `extraPaths` (and verify `expandTilde` coverage for `path` and `cwd`):
```typescript
if (settings.extraPaths?.length > 0) {
    settings.extraPaths = resolveVariables(settings.extraPaths, workspace).map(expandTilde);
}
```

[verified]

edvilme: The shared _getWorkspaceSettings already handles this — lines 210-213 of the shared settings.ts check for extraPaths in settingsDefaults and call resolveVariables(getExtraPaths(namespace, workspace), workspace), which includes expandTilde (line 78). Similarly, cwd gets tilde expansion at lines 216-218, and path/args are resolved at lines 198-199. The only setting that needs post-processing in the wrapper is ignorePatterns, because it goes through the raw settingsDefaults loop (config.get without resolveVariables).

src/common/settings.ts
```diff
@@ -1,204 +1,75 @@
...
-    workspaceSetting.extraPaths = workspaceSetting.extraPaths.map(expandTilde);
-    if (workspaceSetting.cwd.startsWith('~')) {
-        workspaceSetting.cwd = expandTilde(workspaceSetting.cwd);
+    const resolveInterpreter = includeInterpreter ? getInterpreterDetails : undefined;
+    const settings = (await _getWorkspaceSettings(
+        namespace,
+        workspace,
+        FLAKE8_TOOL_CONFIG,
+        resolveInterpreter,
+    )) as ISettings;
```
rchiodo, line 40: Copilot generated:
**[Behavioral regression]** `extraPaths` is listed in `FLAKE8_TOOL_CONFIG.settingsDefaults` alongside `ignorePatterns`, yet only `ignorePatterns` receives `resolveVariables` post-processing. A user configuring `"flake8.extraPaths": ["${workspaceFolder}/stubs"]` will get a literal unresolved string on PYTHONPATH.

Add the same post-processing:
```typescript
if (settings.extraPaths?.length > 0) {
    settings.extraPaths = resolveVariables(settings.extraPaths, workspace);
}
```

Also: the old code applied `expandTilde` to `path`, `extraPaths`, and `cwd`. Import `expandTilde` from `'./envFile'` and add:
```typescript
settings.path = settings.path.map(expandTilde);
settings.extraPaths = settings.extraPaths.map(expandTilde);
if (settings.cwd.startsWith('~')) {
    settings.cwd = expandTilde(settings.cwd);
}
```

[verified]

edvilme: This is already handled by the shared package. Looking at the installed `@vscode/common-python-lsp@0.2.1` (node_modules/@vscode/common-python-lsp/dist/settings.js):

- **extraPaths** (L213-214): `if ('extraPaths' in toolConfig.settingsDefaults) { settings.extraPaths = resolveVariables(getExtraPaths(namespace, workspace), workspace); }` — this overrides the raw value from the settingsDefaults loop with the properly resolved one.
- **path** (L203): `resolveVariables(config.get('path', []), workspace, interpreter)` 
- **cwd** (L218): explicit `expandTilde(settings.cwd)`
- **resolveVariables** (L102): calls `expandTilde(s)` on every entry

The only setting that needs post-processing in the wrapper is `ignorePatterns`, because it only goes through the generic settingsDefaults loop (`config.get` without any resolution) and there is no special handling for it in the shared code (unlike extraPaths which has dedicated logic at L213-214).

rchiodo (approved):
Approved via Review Center.

PR #500, opened 2026-04-28
refactor: delegate logging, CWD, and execution to ToolServer

edvilme:
## Summary

Replaces inline implementations of logging, CWD resolution, tool execution, settings management, and lifecycle cleanup in `lsp_server.py` with calls to the shared `ToolServer` from `@vscode/common-python-lsp` (v0.3.0).

This is the next step after PR #487 which wired the shared package — now we consume the server-level abstractions.

## Changes

- Create `ToolServerConfig` + `ToolServer` instance wrapping the existing `LanguageServer` (with notebook sync)
- Derive `TOOL_MODULE`/`TOOL_DISPLAY`/`TOOL_ARGS`/`MIN_VERSION` from config
- **Delete `get_cwd()` body** (~55 lines) → delegate to `tool_server.get_cwd()`
- **Delete `_to_run_result_with_logging()`** (~9 lines) → handled internally by `execute_tool()`
- **Simplify `_run_tool_on_document()`** (~30 lines saved) → determine mode + build argv, then call `tool_server.execute_tool()` instead of inlining 3 execution strategies
- **Simplify `_run_tool()`** → same delegation pattern for version checks
- **Delete settings functions** (~90 lines) → delegate to `tool_server.get_global_defaults()`, `tool_server.update_workspace_settings()`, `tool_server.get_settings_by_document()`
- **Replace logging functions** (`log_to_output`/`log_error`/`log_warning`/`log_always`) with thin wrappers delegating to `tool_server.log_*()`
- **Replace `on_exit`/`on_shutdown`** bodies with `tool_server.handle_exit()`/`handle_shutdown()`
- **Remove unused `lsp_jsonrpc` import** (execution now handled by ToolServer)
- **Remove dead code**: `_get_document_key()`, `_get_settings_by_path()`

## Backward compatibility

Module-level names `WORKSPACE_SETTINGS`, `GLOBAL_SETTINGS`, `LSP_SERVER`, `_get_global_defaults`, `_get_settings_by_document`, `log_to_output`, `get_cwd` are kept as aliases/thin wrappers so existing tests continue to work unchanged.

## Impact

- **Net: -220 lines** (1062 → 842)
- All 28 TypeScript tests pass
- All 111 Python tests pass
- Lint (flake8 + black + isort + eslint) passes
- Webpack build succeeds

### DIFF
--- a/bundled/tool/lsp_server.py
+++ b/bundled/tool/lsp_server.py
@@ -68,13 +68,13 @@ update_environ_path()
 # **********************************************************
 # Imports needed for the language server goes below this.
 # **********************************************************
-import lsp_jsonrpc as jsonrpc
 import lsp_notebook as notebook
 import lsp_utils as utils
 from lsprotocol import types as lsp
 from pygls import uris
 from pygls.lsp.server import LanguageServer
 from pygls.workspace import TextDocument
+from vscode_common_python_lsp.server import ToolServer, ToolServerConfig
 
 WORKSPACE_SETTINGS = {}
 GLOBAL_SETTINGS = {}
@@ -82,12 +82,37 @@ RUNNER = pathlib.Path(__file__).parent / "lsp_runner.py"
 
 MAX_WORKERS = 5
 
+# Create the LSP server with notebook sync, then wrap in ToolServer for
+# shared logging, CWD resolution, and tool execution.  Settings management
+# remains local because flake8 uses normalize_path(resolve_symlinks=False).
 LSP_SERVER = LanguageServer(
     name="flake8-server",
     version="v0.1.0",
     max_workers=MAX_WORKERS,
     notebook_document_sync=notebook.NOTEBOOK_SYNC_OPTIONS,
 )
+
+FLAKE8_CONFIG = ToolServerConfig(
+    tool_module="flake8",
+    tool_display="Flake8",
+    tool_args=["--format='%(row)d,%(col)d,%(code).1s,%(code)s:%(text)s'"],
+    min_version="5.0.0",
+    runner_script=str(RUNNER),
+    default_notification_level="onError",
+    default_settings={
+        "enabled": True,
+        "severity": {
+            "E": "Error",
+            "F": "Error",
+            "I": "Information",
+            "W": "Warning",
+        },
+        "ignorePatterns": [],
+        "extraPaths": [],
+    },
+)
+
+tool_server = ToolServer(FLAKE8_CONFIG, server=LSP_SERVER)
 
 
 def _get_document_path(document: str) -> str:
@@ -115,16 +140,12 @@ def _get_document_path(document: str) -> str:
 
 
 # **********************************************************
-# Tool specific code goes below this.
+# Tool constants — derived from FLAKE8_CONFIG for backward compatibility.
 # **********************************************************
-TOOL_MODULE = "flake8"
-TOOL_DISPLAY = "Flake8"
-
-# Default arguments always passed to flake8.
-TOOL_ARGS = ["--format='%(row)d,%(col)d,%(code).1s,%(code)s:%(text)s'"]
-
-# Minimum version of flake8 supported.
-MIN_VERSION = "5.0.0"
+TOOL_MODULE = FLAKE8_CONFIG.tool_module
+TOOL_DISPLAY = FLAKE8_CONFIG.tool_display
+TOOL_ARGS = FLAKE8_CONFIG.tool_args
+MIN_VERSION = FLAKE8_CONFIG.min_version
 
 # **********************************************************
 # Linting features start here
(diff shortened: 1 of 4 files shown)


bundled/tool/lsp_server.py
```diff
@@ -882,177 +850,78 @@ def _run_tool_on_document(
     else:
```
rchiodo, line 761: Copilot generated:
**[unverified] High — `sys.path` preservation may be lost in module mode.**

The old code explicitly wrapped module-mode execution in `with utils.substitute_attr(sys, "path", [""] + sys.path[:])` to prevent flake8 from corrupting the interpreter's module search path across invocations. This guard is now deleted — the assumption is that `ToolServer.execute_tool(mode="module")` replicates it internally.

Both the Advocate and Skeptic flag this as the highest-risk item. If the shared package doesn't replicate this guard, 2nd+ invocations in the same server process will see progressively corrupted `sys.path`, causing incorrect module resolution or `ModuleNotFoundError`.

**Action:** Confirm `ToolServer.execute_tool(mode="module")` wraps execution in the equivalent `substitute_attr(sys, "path", ...)` pattern. If it doesn't, this is a regression.

[verified]

edvilme: Confirmed — `ToolServer.execute_tool(mode=\"module\")` wraps execution in the same `substitute_attr(sys, \"path\", [\"\"] + sys.path[:])` guard at [server.py L371](https://github.com/microsoft/vscode-common-python-lsp/blob/ceb9077/python/vscode_common_python_lsp/server.py#L371). No regression — the shared package replicates the exact same pattern.

bundled/tool/lsp_server.py
```diff
@@ -882,177 +850,78 @@ def _run_tool_on_document(
...
-        if result.stderr:
-            log_to_output(result.stderr)
+    result = tool_server.execute_tool(
+        argv=argv,
+        mode=mode,
+        settings=settings,
+        use_stdin=True,
+        cwd=cwd,
+        workspace=code_workspace,
+    )
```
rchiodo, line 806: Copilot generated:
`[unverified]` **Likely double-logging stdout.** `_run_tool` calls `log_to_output(f"\r\n{result.stdout}\r\n")` after `execute_tool()` returns. If `execute_tool()` logs stdout internally (as the old code did in each branch), this duplicates output in the channel. Note that `_run_tool_on_document` does NOT have this trailing log — the inconsistency suggests this is unintentional. Consider removing this line or guarding with `if not tool_server.logs_stdout_internally`.

[verified]

edvilme: Not a double-log — `execute_tool()` does **not** log stdout internally. Looking at [`ToolServer.execute_tool()`](https://github.com/microsoft/vscode-common-python-lsp/blob/ceb9077/python/vscode_common_python_lsp/server.py#L293-L392):

- **path mode** (L336-347): logs `argv`, `CWD`, `stderr` only
- **rpc mode** (L349-366): logs `argv`, `CWD`, `stderr`/`exception` via `_rpc_to_run_result`
- **module mode** (L368-384): logs `argv`, `CWD`, `stderr` only, `exception` on error

None of the three modes log `result.stdout`.

The `log_to_output` at L897 is the **only** place stdout is logged for `_run_tool` — identical to the [old code](https://github.com/microsoft/vscode-flake8/blob/489564e/bundled/tool/lsp_server.py#L974) which also had this line at the end of `_run_tool`.

The inconsistency with `_run_tool_on_document` (no trailing stdout log) is also preserved from the old code — `_run_tool_on_document` never logged stdout at the end; instead, callers like `_linting_helper` and `_lint_notebook_document` log it themselves when publishing diagnostics.

edvilme:
Depends on https://github.com/microsoft/vscode-common-python-lsp/pull/24

bundled/tool/lsp_server.py
```diff
@@ -68,27 +68,54 @@ def update_environ_path() -> None:
...
     notebook_document_sync=notebook.NOTEBOOK_SYNC_OPTIONS,
 )
 
+FLAKE8_CONFIG = ToolServerConfig(
+    tool_module="flake8",
+    tool_display="Flake8",
+    tool_args=["--format='%(row)d,%(col)d,%(code).1s,%(code)s:%(text)s'"],
+    min_version="5.0.0",
+    runner_script=str(RUNNER),
+    default_notification_level="onError",
```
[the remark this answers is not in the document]
edvilme: Not a regression — \default_notification_level=\onError\\ matches the original code exactly:

- Original Python: \GLOBAL_SETTINGS.get(\showNotifications\, \onError\)\
- \package.json\: \lake8.showNotifications\ default is \\onError\\

The TS client always sets \LS_SHOW_NOTIFICATION\ from the VS Code setting (which defaults to \\onError\\ per \package.json\). The \\off\\ fallback in \os.getenv(\LS_SHOW_NOTIFICATION\, \off\)\ only applies if the env var is never set (which doesn't happen in normal operation since the TS server launcher always sets it).

bundled/tool/lsp_server.py
```diff
@@ -647,108 +672,15 @@ def _log_version_info() -> None:
...
-        for _key, settings in WORKSPACE_SETTINGS.items():
-            if settings["workspaceFS"] == str_file_path:
-                return settings
-        file_path = file_path.parent
-
-    setting_values = list(WORKSPACE_SETTINGS.values())
-    if not setting_values:
-        return {}
-    return setting_values[0]
+    return tool_server.get_settings_by_document(document)
```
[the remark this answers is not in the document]
edvilme: This server targets Python 3.10+ (the \python-requires\ in setup and CI matrix both enforce >=3.10). PEP 604 union syntax (\X | None\) is fully supported at runtime on 3.10+.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #528, opened 2026-06-09
refactor: use shared logLegacySettings from common package

edvilme:
## Summary
Reduce code duplication between this extension and the shared package `@vscode/common-python-lsp` (`vscode-common-python-lsp`).

### Changes
- **`src/common/settings.ts`**: Use shared `logLegacySettings()` instead of inline legacy-setting migration warnings

### Impact
- Reduces duplicated code for legacy settings handling
- No behavioral changes — all existing functionality preserved
- No test modifications


rchiodo (approved):
Approved via Review Center.

PR #539, opened 2026-06-23
Upgrade Node to 24.17.0 and give common-python-lsp standalone Dependabot PRs

edvilme:
## Summary
- Pin `NODE_VERSION` to `24.17.0` (latest Node 24 LTS, matching VS Code's bundled runtime) in `pr-check.yml` / `push-check.yml`, and set `.nvmrc` to match.
- Exclude `@vscode/common-python-lsp` (npm) and `vscode-common-python-lsp` (pip) from the weekly grouped Dependabot updates so they get their own standalone PRs.

`@types/node` is unchanged (`22.x`). Dependabot schedules remain weekly.


.github/workflows/pr-check.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: 22.21.1
+  NODE_VERSION: 24.17.0
```
rchiodo, line 7: 📍 .github/workflows/pr-check.yml:7
The Rules reviewer flagged a documented repo instruction violation: "Do not hard-code a Node version… read it from `.nvmrc`." This PR adds `.nvmrc` (correct) but then hard-codes `NODE_VERSION: 24.17.0` here, and the Architect/Skeptic independently flagged that the version is now triplicated (`.nvmrc` + both workflows) and can silently drift on the next bump. Replace the `NODE_VERSION` env + `node-version: ${{ env.NODE_VERSION }}` with `node-version-file: .nvmrc` on the `actions/setup-node` step so `.nvmrc` is the single source of truth. As long as the instruction stands, this cannot be waived for this PR.

[verified]

.github/workflows/push-check.yml
```diff
@@ -9,7 +9,7 @@ on:
       - 'release-*'
 
 env:
-  NODE_VERSION: 22.21.1
+  NODE_VERSION: 24.17.0
```
rchiodo, line 12: 📍 .github/workflows/push-check.yml:12
Same review-rule violation as `pr-check.yml`: `NODE_VERSION` is hard-coded instead of read from the newly-added `.nvmrc`. Switch the `actions/setup-node` step to `node-version-file: .nvmrc` and drop the redundant env literal.

[verified]

.github/dependabot.yml
```diff
@@ -36,6 +39,9 @@ updates:
       pip-minor-patch:
         patterns:
           - '*'
+        # Exclude vscode-common-python-lsp so it gets its own standalone PR (weekly).
+        exclude-patterns:
+          - 'vscode-common-python-lsp'
```
rchiodo, line 44: 📍 .github/dependabot.yml:42
The Skeptic structurally confirmed (by reading every pip manifest) that `vscode-common-python-lsp` is not declared in any file Dependabot's pip ecosystem scans — it exists only as a hardcoded `session.install("vscode-common-python-lsp==0.3.0")` string in `noxfile.py`. Dependabot never tracked it, so this `exclude-patterns` entry is a no-op and produces no standalone PR; the comment ("so it gets its own standalone PR (weekly)") is misleading. Either declare the dependency in a scanned manifest (e.g. `requirements.in` + regenerate `requirements.txt`, and install from it in `noxfile.py`) to get real coverage, or drop this pattern+comment and note that the library is updated manually via `noxfile.py`. The npm-side exclusion is unaffected and works correctly.

[verified]

rchiodo:
Functionally sound. The main theme of feedback: make `.nvmrc` the single source of truth via `node-version-file: .nvmrc` in both workflows (the repo instruction explicitly forbids hard-coding the Node version), and verify the `vscode-common-python-lsp` Dependabot exclusion actually corresponds to a scanned pip manifest.

rchiodo (approved):
Approved via Review Center.

PR #547, opened 2026-06-30
Adopt vscode-common-python-lsp as a git submodule, synced via repository_dispatch, and stop Dependabot duplicates

edvilme:
## Summary

Consumes the shared LSP package (`vscode-common-python-lsp`) as a **git submodule** pinned to a release tag, instead of the published npm/PyPI packages, and keeps that submodule updated via a `repository_dispatch` event from `vscode-common-python-lsp` rather than Dependabot.

### 1. Git submodule
- Adds `external/vscode-common-python-lsp` (`.gitmodules`) pinned to a shared-package release tag.
- **TypeScript:** `package.json` sources it via `"@vscode/common-python-lsp": "file:external/vscode-common-python-lsp/typescript"`; `build/postinstall.js` builds the submodule's TS output on install (skips when already built or when the submodule toolchain is absent).
- **Python:** `noxfile.py` installs `./external/vscode-common-python-lsp/python` into `./bundled/libs` with `--no-deps --upgrade`, so the bundled copy matches the pinned submodule commit.

### 2. New `.github/workflows/shared-package-submodule-sync.yml`
Handles the `shared-package-release` `repository_dispatch`: branches off `main`, advances the submodule pointer to the released tag, refreshes the npm lockfile, verifies the pinned tag and the Python `requires-python` floor, then **pushes the branch and opens a tracking issue** with a manual compare/PR link (org settings prevent the workflow from opening PRs automatically).

### 3. `.github/dependabot.yml`
Moves the shared package into the npm and pip `ignore` lists so Dependabot no longer opens duplicate update PRs for it.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
 
       - name: Build VSIX
         uses: ./.github/actions/build-vsix
@@ -31,6 +33,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
 
       - name: Lint
         uses: ./.github/actions/lint
@@ -54,6 +58,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          submodules: recursive
           path: ${{ env.special-working-directory-relative }}
 
       # Install bundled libs using env.PYTHON_VERSION even though you test it on other versions.
@@ -106,6 +111,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          submodules: recursive
           path: ${{ env.special-working-directory-relative }}
 
       - name: Use Node.js ${{ env.NODE_VERSION }}
--- a/.github/workflows/shared-package-release.yml
+++ b/.github/workflows/shared-package-release.yml
@@ -8,9 +8,15 @@ permissions:
   contents: write
   issues: write
 
+concurrency:
+  group: 'shared-package-release-${{ github.event.client_payload.release_tag }}'
+  cancel-in-progress: false
+
 env:
   NODE_VERSION: 24.17.0
-  PYTHON_VERSION: '3.x'
+  # Match the Python version documented in requirements.in so the regenerated
+  # lockfile stays valid across the supported Python range.
+  PYTHON_VERSION: '3.9'
 
 jobs:
   update-shared-packages:
@@ -30,6 +36,20 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 
+      - name: Validate dispatch payload
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          if [ -z "${RELEASE_TAG}" ]; then
+            echo 'release_tag is missing from the dispatch payload.' >&2
+            exit 1
+          fi
+          if ! printf '%s' "${RELEASE_TAG}" | grep -Eq '^v?[0-9A-Za-z][0-9A-Za-z.+-]*$'; then
+            echo "release_tag '${RELEASE_TAG}' is not a valid version/ref." >&2
+            exit 1
+          fi
+
       - name: Normalize release tag
         id: release_version
         env:
@@ -46,6 +66,8 @@ jobs:
           set -euo pipefail
           BRANCH="shared-package-v${RELEASE_TAG#v}"
           git fetch origin main
+          # Fetch any pre-existing remote branch so --force-with-lease has a ref.
+          git fetch origin "+refs/heads/${BRANCH}:refs/remotes/origin/${BRANCH}" 2>/dev/null || true
           git checkout -B "$BRANCH" origin/main
 
       - name: Update npm dependency
@@ -75,10 +97,20 @@ jobs:
             exit 0
           fi
           python -m pip install --upgrade uv
-          sed -i -E "s/^${PIP_DEP}([<>=!~].*)?\$/${PIP_DEP}==${RELEASE_VERSION}/" requirements.in
-          uv pip compile --generate-hashes --upgrade -o requirements.txt requirements.in
+          # Rewrite the pin (case-insensitive) for the shared package only.
+          sed -i -E "s/^${PIP_DEP}([<>=!~].*)?\$/${PIP_DEP}==${RELEASE_VERSION}/I" requirements.in
+          if ! grep -iEq "^${PIP_DEP}==${RELEASE_VERSION}([[:space:];].*)?\$" requirements.in; then
+            echo "Failed to pin ${PIP_DEP}==${RELEASE_VERSION} in requirements.in." >&2
+            exit 1
+          fi
+          # Scope the re-resolve to the shared package and pin to the documented
+          # Python version so unrelated deps and version markers are preserved.
+          uv pip compile --generate-hashes \
+            --upgrade-package "${PIP_DEP}" \
+            --python-version "${PYTHON_VERSION}" \
+            -o requirements.txt requirements.in
 
-      - name: Commit branch and open tracking issue if changed
+      - name: Push branch and open tracking issue if changed
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
           RELEASE_URL: ${{ github.event.client_payload.release_url }}
@@ -95,8 +127,18 @@ jobs:
           BRANCH="shared-package-v${RELEASE_TAG#v}"
           git add -A
           git commit -m "Upgrade shared package to ${RELEASE_TAG}"
-          git push --set-upstream origin "$BRANCH"
+          git push --force-with-lease --set-upstream origin "$BRANCH"
           COMPARE_URL="https://github.com/${REPO}/compare/main...${BRANCH}?expand=1"
+          TITLE="[Shared Package] Open PR to upgrade to ${RELEASE_TAG}"
+          {
+            echo "### Shared package upgrade ready"
+            echo ''
+            echo "Branch \`${BRANCH}\` has been pushed."
+            echo ''
+            echo "[Open the pull request](${COMPARE_URL})"
+            echo ''
+            echo "Source release: ${RELEASE_URL}"
+          } >> "$GITHUB_STEP_SUMMARY"
           BODY_FILE="$(mktemp)"
           cat > "$BODY_FILE" <<EOF
           Branch \`${BRANCH}\` has been pushed with the shared package upgrade to ${RELEASE_TAG}.
@@ -107,6 +149,13 @@ jobs:
 
           Source release: ${RELEASE_URL}
           EOF
-          gh issue create \
-            --title "[Shared Package] Open PR to upgrade to ${RELEASE_TAG}" \
-            --body-file "$BODY_FILE"
+          # Reuse an existing open tracking issue for this release if present.
+          EXISTING="$(gh issue list --repo "$REPO" --state open --search "in:title ${TITLE}" --json number,title --jq "map(select(.title == \"${TITLE}\")) | .[0].number // empty" 2>/dev/null || true)"
+          if [ -n "${EXISTING}" ]; then
+            echo "Reusing existing tracking issue #${EXISTING}."
+            gh issue comment "${EXISTING}" --repo "$REPO" --body-file "$BODY_FILE" \
+              || echo 'Could not comment on issue; compare URL is in the job summary.'
+          else
+            gh issue create --repo "$REPO" --title "${TITLE}" --body-file "$BODY_FILE" \
+              || echo 'Could not create issue (Issues may be disabled); compare URL is in the job summary.'
+          fi
--- a/.github/workflows/shared-package-release.yml
+++ b/.github/workflows/shared-package-release.yml
@@ -75,25 +75,8 @@ jobs:
             exit 0
           fi
           python -m pip install --upgrade uv
-          python - <<'PY'
-          from pathlib import Path
-          import os
-          import re
-
-          dep = os.environ["PIP_DEP"]
-          version = os.environ["RELEASE_VERSION"]
-          pinned = f"{dep}=={version}"
-          req_in = Path("requirements.in")
-          txt = req_in.read_text(encoding="utf-8")
-          updated = re.sub(
-              rf"^{re.escape(dep)}([<>=!~].*)?$",
-              pinned,
-              txt,
-              flags=re.MULTILINE,
-          )
-          req_in.write_text(updated, encoding="utf-8")
-          PY
-          uv pip compile --generate-hashes ./requirements.in -o requirements.txt
+          sed -i -E "s/^${PIP_DEP}([<>=!~].*)?\$/${PIP_DEP}==${RELEASE_VERSION}/" requirements.in
+          uv pip compile --generate-hashes --upgrade -o requirements.txt requirements.in
 
       - name: Commit branch and open tracking issue if changed
         env:
--- a/.github/workflows/shared-package-release.yml
+++ b/.github/workflows/shared-package-release.yml
@@ -1,0 +1,129 @@
+name: Shared Package Release Handler
+
+on:
+  repository_dispatch:
+    types: [shared-package-release]
+
+permissions:
+  contents: write
+  issues: write
+
+env:
+  NODE_VERSION: 24.17.0
+  PYTHON_VERSION: '3.x'
+
+jobs:
+  update-shared-packages:
+    name: Update shared packages
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install Node
+        uses: actions/setup-node@v6
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Normalize release tag
+        id: release_version
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          VERSION="${RELEASE_TAG#v}"
+          echo "value=$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Create branch for this release
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          BRANCH="shared-package-v${RELEASE_TAG#v}"
+          git fetch origin main
+          git checkout -B "$BRANCH" origin/main
+
+      - name: Update npm dependency
+        env:
+          NPM_DEP: ${{ github.event.client_payload.npm_dependency }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
+          set -euo pipefail
+          if [ -z "${NPM_DEP}" ]; then
+            echo 'No npm dependency in payload; skipping.'
+            exit 0
+          fi
+          npm install "${NPM_DEP}@${RELEASE_VERSION}"
+
+      - name: Update pip dependency
+        env:
+          PIP_DEP: ${{ github.event.client_payload.pip_dependency }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
+          set -euo pipefail
+          if [ -z "${PIP_DEP}" ]; then
+            echo 'No pip dependency in payload; skipping.'
+            exit 0
+          fi
+          if [ ! -f requirements.in ]; then
+            echo 'No requirements.in found; skipping.'
+            exit 0
+          fi
+          python -m pip install --upgrade uv
+          python - <<'PY'
+          from pathlib import Path
+          import os
+          import re
+
+          dep = os.environ["PIP_DEP"]
+          version = os.environ["RELEASE_VERSION"]
+          pinned = f"{dep}=={version}"
+          req_in = Path("requirements.in")
+          txt = req_in.read_text(encoding="utf-8")
+          updated = re.sub(
+              rf"^{re.escape(dep)}([<>=!~].*)?$",
+              pinned,
+              txt,
+              flags=re.MULTILINE,
+          )
+          req_in.write_text(updated, encoding="utf-8")
+          PY
+          uv pip compile --generate-hashes ./requirements.in -o requirements.txt
+
+      - name: Commit branch and open tracking issue if changed
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+          RELEASE_URL: ${{ github.event.client_payload.release_url }}
+          REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          if git diff --quiet; then
+            echo 'No changes detected.'
+            exit 0
+          fi
+          BRANCH="shared-package-v${RELEASE_TAG#v}"
+          git add -A
+          git commit -m "Upgrade shared package to ${RELEASE_TAG}"
+          git push --set-upstream origin "$BRANCH"
+          COMPARE_URL="https://github.com/${REPO}/compare/main...${BRANCH}?expand=1"
+          BODY_FILE="$(mktemp)"
+          cat > "$BODY_FILE" <<EOF
+          Branch \`${BRANCH}\` has been pushed with the shared package upgrade to ${RELEASE_TAG}.
+
+          Organization settings prevent this workflow from opening pull requests automatically. Please open the PR manually:
+
+          ${COMPARE_URL}
+
+          Source release: ${RELEASE_URL}
+          EOF
+          gh issue create \
+            --title "[Shared Package] Open PR to upgrade to ${RELEASE_TAG}" \
+            --body-file "$BODY_FILE"
--- a/.github/workflows/shared-package-submodule-sync.yml
+++ b/.github/workflows/shared-package-submodule-sync.yml
@@ -1,4 +1,4 @@
-name: Shared Package Release Handler
+name: Shared Package Submodule Sync
 
 on:
   repository_dispatch:
@@ -9,32 +9,22 @@ permissions:
   issues: write
 
 concurrency:
-  group: 'shared-package-release-${{ github.event.client_payload.release_tag }}'
+  group: 'shared-package-submodule-${{ github.event.client_payload.release_tag }}'
   cancel-in-progress: false
 
 env:
-  NODE_VERSION: 24.17.0
-  # Match the Python version documented in requirements.in so the regenerated
-  # lockfile stays valid across the supported Python range.
-  PYTHON_VERSION: '3.9'
+  SUBMODULE_PATH: external/vscode-common-python-lsp
 
 jobs:
-  update-shared-packages:
-    name: Update shared packages
+  update-submodule:
+    name: Update shared package submodule
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v6
-
-      - name: Install Node
-        uses: actions/setup-node@v6
         with:
-          node-version: ${{ env.NODE_VERSION }}
-
-      - name: Install Python
-        uses: actions/setup-python@v6
-        with:
-          python-version: ${{ env.PYTHON_VERSION }}
+          submodules: recursive
+          fetch-depth: 0
 
       - name: Validate dispatch payload
         env:
@@ -50,15 +40,6 @@ jobs:
             exit 1
           fi
 
-      - name: Normalize release tag
-        id: release_version
-        env:
-          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
-        run: |
-          set -euo pipefail
-          VERSION="${RELEASE_TAG#v}"
-          echo "value=$VERSION" >> "$GITHUB_OUTPUT"
-
       - name: Create branch for this release
         env:
           RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
@@ -69,46 +50,28 @@ jobs:
           # Fetch any pre-existing remote branch so --force-with-lease has a ref.
           git fetch origin "+refs/heads/${BRANCH}:refs/remotes/origin/${BRANCH}" 2>/dev/null || true
           git checkout -B "$BRANCH" origin/main
+          # Re-sync the submodule to whatever the freshly checked-out main records.
+          git submodule update --init --recursive "${SUBMODULE_PATH}"
 
-      - name: Update npm dependency
+      - name: Move submodule to the released commit
         env:
-          NPM_DEP: ${{ github.event.client_payload.npm_dependency }}
-          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
         run: |
           set -euo pipefail
-          if [ -z "${NPM_DEP}" ]; then
-            echo 'No npm dependency in payload; skipping.'
-            exit 0
+          cd "${SUBMODULE_PATH}"
+          git fetch --tags --force origin
+          STRIPPED="${RELEASE_TAG#v}"
+          if git rev-parse -q --verify "refs/tags/${RELEASE_TAG}^{commit}" >/dev/null; then
+            TARGET="refs/tags/${RELEASE_TAG}"
+          elif git rev-parse -q --verify "refs/tags/v${STRIPPED}^{commit}" >/dev/null; then
+            TARGET="refs/tags/v${STRIPPED}"
+          else
+            echo "Release tag '${RELEASE_TAG}' not found upstream; falling back to origin/main." >&2
+            git fetch origin main
+            TARGET="origin/main"
           fi
-          npm install "${NPM_DEP}@${RELEASE_VERSION}"
-
-      - name: Update pip dependency
-        env:
-          PIP_DEP: ${{ github.event.client_payload.pip_dependency }}
-          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
-        run: |
-          set -euo pipefail
-          if [ -z "${PIP_DEP}" ]; then
-            echo 'No pip dependency in payload; skipping.'
-            exit 0
-          fi
-          if [ ! -f requirements.in ]; then
-            echo 'No requirements.in found; skipping.'
-            exit 0
-          fi
-          python -m pip install --upgrade uv
-          # Rewrite the pin (case-insensitive) for the shared package only.
-          sed -i -E "s/^${PIP_DEP}([<>=!~].*)?\$/${PIP_DEP}==${RELEASE_VERSION}/I" requirements.in
-          if ! grep -iEq "^${PIP_DEP}==${RELEASE_VERSION}([[:space:];].*)?\$" requirements.in; then
-            echo "Failed to pin ${PIP_DEP}==${RELEASE_VERSION} in requirements.in." >&2
-            exit 1
-          fi
-          # Scope the re-resolve to the shared package and pin to the documented
-          # Python version so unrelated deps and version markers are preserved.
-          uv pip compile --generate-hashes \
-            --upgrade-package "${PIP_DEP}" \
-            --python-version "${PYTHON_VERSION}" \
-            -o requirements.txt requirements.in
+          echo "Checking out submodule at ${TARGET}."
+          git checkout --detach "${TARGET}"
 
       - name: Push branch and open tracking issue if changed
         env:
@@ -120,18 +83,19 @@ jobs:
           set -euo pipefail
           git config user.name 'github-actions[bot]'
           git config user.email 'github-actions[bot]@users.noreply.github.com'
-          if git diff --quiet; then
-            echo 'No changes detected.'
+          # Only the submodule gitlink should have moved.
+          if git diff --quiet -- "${SUBMODULE_PATH}"; then
+            echo 'Submodule already at the released commit; nothing to do.'
             exit 0
           fi
           BRANCH="shared-package-v${RELEASE_TAG#v}"
-          git add -A
-          git commit -m "Upgrade shared package to ${RELEASE_TAG}"
+          git add "${SUBMODULE_PATH}"
+          git commit -m "Update shared package submodule to ${RELEASE_TAG}"
           git push --force-with-lease --set-upstream origin "$BRANCH"
           COMPARE_URL="https://github.com/${REPO}/compare/main...${BRANCH}?expand=1"
-          TITLE="[Shared Package] Open PR to upgrade to ${RELEASE_TAG}"
+          TITLE="[Shared Package] Open PR to update submodule to ${RELEASE_TAG}"
           {
-            echo "### Shared package upgrade ready"
+            echo "### Shared package submodule update ready"
             echo ''
             echo "Branch \`${BRANCH}\` has been pushed."
             echo ''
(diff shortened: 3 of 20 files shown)


.github/workflows/shared-package-release.yml
```diff
@@ -0,0 +1,112 @@
...
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
+          set -euo pipefail
+          if [ -z "${PIP_DEP}" ]; then
+            echo 'No pip dependency in payload; skipping.'
+            exit 0
+          fi
+          if [ ! -f requirements.in ]; then
+            echo 'No requirements.in found; skipping.'
+            exit 0
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:74
`uv pip compile --generate-hashes --upgrade` re-resolves **all** deps to latest, not just the shared package. `requirements.in` also pins `pygls`, `packaging`, and `flake8`, so a shared-package release silently drags unrelated upgrades into a branch/issue titled "Upgrade shared package to vX" (Skeptic: High, Architect: High; the Advocate could find no evidence the broad sweep is intentional). Scope it with `--upgrade-package "${PIP_DEP}"` so only the shared package and its forced transitives move.

[verified]

.github/workflows/shared-package-release.yml
```diff
@@ -0,0 +1,112 @@
...
+        env:
+          PIP_DEP: ${{ github.event.client_payload.pip_dependency }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
+          set -euo pipefail
+          if [ -z "${PIP_DEP}" ]; then
+            echo 'No pip dependency in payload; skipping.'
+            exit 0
+          fi
+          if [ ! -f requirements.in ]; then
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:73
`sed` exits 0 even when it substitutes nothing, so a non-matching line (PyPI-normalized name, extras like `[foo]`, or an environment marker such as `; python_version >= "3.9"`) leaves the pin untouched. Combined with `--upgrade` still rewriting `requirements.txt`, `git diff --quiet` then sees changes and a commit + tracking issue is created claiming an upgrade that never touched the shared pin — a silent integrity gap. After the `sed`, assert it took: `grep -qE "^${PIP_DEP}==${RELEASE_VERSION}\$" requirements.in || { echo 'pip pin not updated'; exit 1; }`.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,112 @@
...
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          if git diff --quiet; then
+            echo 'No changes detected.'
+            exit 0
+          fi
+          BRANCH="shared-package-v${RELEASE_TAG#v}"
+          git add -A
+          git commit -m "Upgrade shared package to ${RELEASE_TAG}"
+          git push --set-upstream origin "$BRANCH"
+          COMPARE_URL="https://github.com/${REPO}/compare/main...${BRANCH}?expand=1"
```
rchiodo, line 251: 📍 .github/workflows/shared-package-release.yml:85
Re-dispatch/re-run is non-idempotent (all three reviewers). "Create branch" does `checkout -B "$BRANCH" origin/main` (resetting to main), then this step does `git push --set-upstream origin "$BRANCH"` with no `--force`. If `shared-package-v<ver>` already exists remotely, the push is rejected as non-fast-forward and the run fails **after** committing — so no tracking issue is created and operators see a red workflow. Separately, `gh issue create` runs unconditionally, so retries spawn duplicate "[Shared Package] Open PR…" issues. Use `git push --force-with-lease` and guard issue creation against an existing open issue/branch.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,112 @@
...
+    types: [shared-package-release]
+
+permissions:
+  contents: write
+  issues: write
+
+env:
+  NODE_VERSION: 24.17.0
+  PYTHON_VERSION: '3.x'
+
```
rchiodo, line 13: 📍 .github/workflows/shared-package-release.yml:13
`PYTHON_VERSION: '3.x'` resolves to the latest Python, but `requirements.in`'s own header documents "Use `uv` with Python 3.9." `--generate-hashes` pins and environment markers can differ across Python versions, so the CI-generated `requirements.txt` may diverge from what a maintainer produces locally (Advocate, Medium; [unverified] — confirming requires compiling under 3.9 vs latest and diffing). Pin `PYTHON_VERSION: '3.9'` to match the documented process.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,112 @@
...
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          BRANCH="shared-package-v${RELEASE_TAG#v}"
+          git fetch origin main
+          git checkout -B "$BRANCH" origin/main
+
+      - name: Update npm dependency
+        env:
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:48
`BRANCH="shared-package-v${RELEASE_TAG#v}"` strips only one leading `v` and does no validation. Since `client_payload` is influenceable by anyone able to dispatch, a tag containing spaces, `..`, `~`, `^`, `:`, or `/` produces an invalid git ref and an opaque `checkout -B`/`push` failure (env-quoting prevents injection, so this is DoS-only). Validate `RELEASE_TAG` early against a strict pattern (e.g. `^v?[0-9A-Za-z.\-]+$`) and fail fast with a clear message.

[verified]

rchiodo:
The dispatch workflow is a reasonable approach, but the dependency-pinning step needs to be made precise and verifiable before merge: scope the pip upgrade to the shared package and assert the pin actually changed. A few retry/idempotency and reproducibility hardening notes are also worth addressing.

edvilme:
Thanks @rchiodo — addressed the feedback in the latest commit:

- **Python version / reproducibility:** the workflow now pins `PYTHON_VERSION` to the version documented in this repo's `requirements.in` and passes `--python-version` to `uv pip compile`, so environment markers (e.g. `exceptiongroup`) aren't dropped when the lockfile is regenerated.
- **Blanket upgrade:** replaced `--upgrade` with `--upgrade-package "${PIP_DEP}"` so only the shared package (and its forced transitives) move.
- **Silent `sed` no-op:** the pin rewrite is now case-insensitive and is followed by a `grep` assertion that fails the job if `requirements.in` wasn't actually re-pinned.
- **Payload validation:** a new step fails fast if `release_tag` is empty or not version/ref-shaped (`^v?[0-9A-Za-z][0-9A-Za-z.+-]*$`), preventing degenerate branches and `npm install pkg@` resolving to latest.
- **Idempotency / re-runs:** added a `concurrency` group keyed on `release_tag`, switched the push to `--force-with-lease` (fetching any existing remote branch first), de-duplicated the tracking issue (reuse/comment on an existing open one), and made issue creation tolerant of Issues being disabled by also writing the compare URL to the job summary.

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,161 @@
...
+      - name: Create branch for this release
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          BRANCH="shared-package-v${RELEASE_TAG#v}"
+          git fetch origin main
+          # Fetch any pre-existing remote branch so --force-with-lease has a ref.
+          git fetch origin "+refs/heads/${BRANCH}:refs/remotes/origin/${BRANCH}" 2>/dev/null || true
+          git checkout -B "$BRANCH" origin/main
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:70
New regression (Skeptic + Architect, both Medium): the `--force-with-lease` added to fix prior finding #3 provides illusory protection. Line 70 re-fetches the current remote tip into `refs/remotes/origin/${BRANCH}` immediately before line 71 resets the local branch to `origin/main`, and the push (L130) leases against that just-refreshed ref — so the lease passes unconditionally, behaving like plain `--force`. If a maintainer pushed fixup commits onto the open PR branch and the same `release_tag` is re-dispatched, those commits are silently clobbered. Either commit *onto* the existing remote branch instead of `checkout -B ... origin/main`, or short-circuit when `refs/remotes/origin/${BRANCH}` already exists and an open PR/issue references it.

[verified]

.github/workflows/shared-package-release.yml
```diff
@@ -0,0 +1,161 @@
...
+            echo 'No requirements.in found; skipping.'
+            exit 0
+          fi
+          python -m pip install --upgrade uv
+          # Rewrite the pin (case-insensitive) for the shared package only.
+          sed -i -E "s/^${PIP_DEP}([<>=!~].*)?\$/${PIP_DEP}==${RELEASE_VERSION}/I" requirements.in
+          if ! grep -iEq "^${PIP_DEP}==${RELEASE_VERSION}([[:space:];].*)?\$" requirements.in; then
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:101
The whole-line `sed` replacement strips any trailing environment marker or inline comment on the pin line (e.g. `vscode-common-python-lsp>=1.0 ; python_version >= "3.9"` becomes bare `...==X`), which can change resolution semantics. The Skeptic notes the new `grep` assertion does not catch this because it *allows but does not require* a trailing marker (`([[:space:];].*)?`), giving false confidence. Preserve any trailing `;`-marker/comment in the replacement, or assert the marker is still present when one was originally there.

[verified]

.github/workflows/shared-package-release.yml
```diff
@@ -0,0 +1,161 @@
...
+
+      - name: Update pip dependency
+        env:
+          PIP_DEP: ${{ github.event.client_payload.pip_dependency }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
+          set -euo pipefail
+          if [ -z "${PIP_DEP}" ]; then
+            echo 'No pip dependency in payload; skipping.'
+            exit 0
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:93
Validation is much improved, but the charset `^v?[0-9A-Za-z][0-9A-Za-z.+-]*$` still admits a few git-invalid refs (`v1..2`, trailing-dot `v1.`) and non-version tags like `release-1.2.3` (only a leading `v` is stripped, so these fail opaquely later at `npm install pkg@<garbage>` / `checkout -B`). Tighten to disallow `\.\.`/trailing `.` and require digits after the optional `v` so bad payloads fail fast here with a clear message.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,161 @@
...
+            echo "Failed to pin ${PIP_DEP}==${RELEASE_VERSION} in requirements.in." >&2
+            exit 1
+          fi
+          # Scope the re-resolve to the shared package and pin to the documented
+          # Python version so unrelated deps and version markers are preserved.
+          uv pip compile --generate-hashes \
+            --upgrade-package "${PIP_DEP}" \
+            --python-version "${PYTHON_VERSION}" \
+            -o requirements.txt requirements.in
+
```
rchiodo, line 151: 📍 .github/workflows/shared-package-release.yml:112
Carried-over SSOT nit (Architect, Low): `BRANCH="shared-package-v${RELEASE_TAG#v}"` is recomputed here and again at the push step (L130), and a separate "Normalize release tag" step already emits `value="${RELEASE_TAG#v}"` — three independent derivations of the same value. Promote `BRANCH` to a step output and reference it everywhere to prevent drift if the naming scheme changes.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,161 @@
...
+      - name: Push branch and open tracking issue if changed
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:115
Minor robustness (Skeptic, Low): `2>/dev/null || true` swallows *all* fetch errors as "branch absent", including transient network/auth failures. If the branch actually exists but the fetch failed, the subsequent `--force-with-lease` expects "ref absent" and the push is rejected non-fast-forward *after* committing, leaving a red run with no tracking issue. Consider distinguishing "branch does not exist" from a real fetch failure.

[verified]

.github/workflows/shared-package-release.yml
```diff
@@ -0,0 +1,161 @@
...
+        env:
+          NPM_DEP: ${{ github.event.client_payload.npm_dependency }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
+          set -euo pipefail
+          if [ -z "${NPM_DEP}" ]; then
+            echo 'No npm dependency in payload; skipping.'
+            exit 0
+          fi
+          npm install "${NPM_DEP}@${RELEASE_VERSION}"
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:81
The change guard uses `git diff --quiet`, which only inspects tracked, modified files. If `npm install` creates an untracked `package-lock.json` where none existed, the guard reports "No changes detected" and exits before `git add -A` would stage it (theoretical for this repo since the lockfile already exists). Prefer `git status --porcelain` to also catch untracked files.

[verified]

.github/workflows/shared-package-submodule-sync.yml
rchiodo, line 151: 📍 .github/workflows/shared-package-release.yml:78
The `REPO` env var is computed and passed but `gh issue create` has no `--repo` flag — it relies on cwd auto-detection. `COMPARE_URL` does use `$REPO`, so the inconsistency is real; this looks like the issue-creation path wasn't exercised end-to-end (Skeptic). Either add `--repo "$REPO"` for robustness or drop the unused env.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,161 @@
...
+            echo 'No npm dependency in payload; skipping.'
+            exit 0
+          fi
+          npm install "${NPM_DEP}@${RELEASE_VERSION}"
+
+      - name: Update pip dependency
+        env:
+          PIP_DEP: ${{ github.event.client_payload.pip_dependency }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value }}
+        run: |
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:46
SSOT: `BRANCH="shared-package-v${RELEASE_TAG#v}"` is recomputed independently in the "Create branch" step and the "Commit" step, and the compare URL depends on it — three consumers, no single definition (Architect, Medium). If the naming scheme changes, the pushed branch and the issue's compare link can drift into a dead link. Emit `branch` from the existing `release_version` step and reference it everywhere.

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,161 @@
...
+      - name: Normalize release tag
+        id: release_version
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          VERSION="${RELEASE_TAG#v}"
+          echo "value=$VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Create branch for this release
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:53
Minor robustness: `PIP_DEP` is interpolated raw into a `sed -E` pattern, so a name containing `.` (valid in PyPI names) is treated as regex any-char; and the npm/pip steps `exit 0` on an empty payload field, silently skipping half the work while still proceeding to commit/issue with no signal that the dispatch was incomplete (Skeptic). Consider escaping the dep name and surfacing a warning on a partial payload rather than silently continuing.

[verified]

.github/workflows/shared-package-release.yml
```diff
@@ -0,0 +1,161 @@
...
+on:
+  repository_dispatch:
+    types: [shared-package-release]
+
+permissions:
+  contents: write
+  issues: write
+
+concurrency:
+  group: 'shared-package-release-${{ github.event.client_payload.release_tag }}'
```
rchiodo: 📍 .github/workflows/shared-package-release.yml:12
`NODE_VERSION: 24.17.0` is hardcoded. [unverified] If this repo carries a `.nvmrc`, this becomes a second source of truth for the Node version; read it from `.nvmrc` instead. Otherwise fine (Architect, Low).

[verified]

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,125 @@
...
+        env:
+          RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+        run: |
+          set -euo pipefail
+          BRANCH="shared-package-v${RELEASE_TAG#v}"
+          git fetch origin main
+          # Fetch any pre-existing remote branch so --force-with-lease has a ref.
+          git fetch origin "+refs/heads/${BRANCH}:refs/remotes/origin/${BRANCH}" 2>/dev/null || true
+          git checkout -B "$BRANCH" origin/main
+          # Re-sync the submodule to whatever the freshly checked-out main records.
```
rchiodo: The "Create branch" step fetches the remote branch tip into `refs/remotes/origin/${BRANCH}` immediately before `git checkout -B "$BRANCH" origin/main`, and the push step runs a bare `git push --force-with-lease`. A bare lease uses `refs/remotes/origin/<branch>` as its baseline — which was just refreshed to the live remote tip — so the lease always passes and degrades to a plain `--force`. If the same `release_tag` is re-dispatched after a maintainer pushed commits onto the branch, those commits are silently clobbered. Fix: commit onto the existing branch (`git checkout -B "$BRANCH" "origin/${BRANCH}"` when it exists) or short-circuit when the branch already exists with an open PR/issue; if overwrite is truly intended, drop the misleading `--force-with-lease`.

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,125 @@
...
+          STRIPPED="${RELEASE_TAG#v}"
+          if git rev-parse -q --verify "refs/tags/${RELEASE_TAG}^{commit}" >/dev/null; then
+            TARGET="refs/tags/${RELEASE_TAG}"
+          elif git rev-parse -q --verify "refs/tags/v${STRIPPED}^{commit}" >/dev/null; then
+            TARGET="refs/tags/v${STRIPPED}"
+          else
+            echo "Release tag '${RELEASE_TAG}' not found upstream; falling back to origin/main." >&2
+            git fetch origin main
+            TARGET="origin/main"
+          fi
```
rchiodo, line 121: Silent-integrity gap: when neither `refs/tags/${RELEASE_TAG}` nor `refs/tags/v${STRIPPED}` resolves (tag-propagation lag, a regex-valid-but-non-release tag, or a deleted upstream tag), the submodule is silently moved to `origin/main` (a moving target). The commit message (`Update shared package submodule to ${RELEASE_TAG}`) and the tracking issue both then claim the release tag while actually pointing at unreleased main HEAD, and the run stays green (only an `echo … >&2`). A reviewer could merge trusting it is the released commit. Fail the run (`exit 1`) when the release tag can't be resolved — a missing tag is an error, not a silent substitution.

package.json
```diff
@@ -54,6 +54,7 @@
     "main": "./dist/extension.js",
     "l10n": "./l10n",
     "scripts": {
+        "postinstall": "npm --prefix external/vscode-common-python-lsp/typescript run build",
```
rchiodo: `"postinstall": "npm --prefix external/vscode-common-python-lsp/typescript run build"` couples every `npm install`/`npm ci` to an initialized submodule. On a fresh clone without `--recurse-submodules`, `external/vscode-common-python-lsp/typescript` won't exist and the install aborts with a non-obvious `npm --prefix <missing>` error. CI is covered because the pipelines add `git submodule update --init --recursive` before `npm ci`, but local dev is not. Guard the postinstall to skip with a clear warning when the submodule dir is absent, and document the recursive-clone requirement.

rchiodo:
The submodule-based sync approach is reasonable, but the new workflow has two integrity/safety issues worth fixing before merge: (1) the bare `--force-with-lease` after re-fetching the remote branch tip provides no real protection and can overwrite maintainer commits, and (2) the tag-resolution step silently falls back to `origin/main` while still committing/labeling as the release tag. Note: the first-pass candidates referencing `.github/workflows/shared-package-release.yml` were dropped because that file is not part of this PR's diff (the added workflow is `shared-package-submodule-sync.yml`).

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,126 @@
...
+            done
+            if [ -z "${TARGET}" ]; then
+              echo "::error::Release tag '${RELEASE_TAG}' was not found in the shared package repository."
+              exit 1
+            fi
+            echo "Checking out submodule at ${TARGET}."
+            git checkout --detach "${TARGET}"
+          )
+
+          # 4. Stop early if the submodule pointer did not move.
```
rchiodo: 📍 .github/workflows/shared-package-submodule-sync.yml:62
**`--force-with-lease` provides no protection here.** Line 63 fetches the live remote tip into `refs/remotes/origin/${BRANCH}`, then `checkout -B "${BRANCH}" "${MAIN_SHA}"` rebuilds the branch from main, and the bare `--force-with-lease` push (line 86) leases against that just-refreshed ref — so it always passes and behaves like plain `--force`. On a same-tag re-dispatch after a maintainer pushes fixups onto the branch, those commits are silently overwritten, and the new `retry()` wrapper makes the destructive push retry-hardened. Fix: commit onto the existing branch (`git checkout -B "${BRANCH}" "origin/${BRANCH}"` when it exists) or short-circuit when the branch already has an open tracking issue; if overwrite is truly intended, drop the misleading `--force-with-lease` (and its comment) and use an honest `--force`.

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,126 @@
+name: Shared Package Submodule Sync
+
+on:
+  repository_dispatch:
+    types: [shared-package-release]
```
rchiodo, line 5: 📍 .github/workflows/shared-package-submodule-sync.yml:1
The PR description narrates a `.github/workflows/shared-package-release.yml` that bumps npm, rewrites `requirements.in`, and recompiles with `uv` — none of which exists in this diff. The actual file is `shared-package-submodule-sync.yml` using git-submodule vendoring. Update the PR description so reviewers don't hunt for the wrong file/behavior.

rchiodo:
Main theme: the submodule-sync workflow's `--force-with-lease` protection is illusory (the branch tip is fetched into the tracking ref right before the push, so the lease always passes). Please make the force-push honest or commit onto the existing branch. Also note the PR description describes a different workflow (`shared-package-release.yml` with npm/uv recompile) than what's actually implemented here (git-submodule vendoring), and the `postinstall` build hook can break non-recursive/local clones.

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,139 @@
...
+
+          # 1. Validate the dispatch payload (require a real dotted version).
+          if ! printf '%s' "${RELEASE_TAG}" | grep -Eq '^v?[0-9]+\.[0-9]+\.[0-9]+([-+.][0-9A-Za-z.+-]+)*$'; then
+            echo "::error::release_tag '${RELEASE_TAG:-<empty>}' is missing or not a valid semver release."
+            exit 1
+          fi
+          VERSION="${RELEASE_TAG#v}"
+          BRANCH="shared-package-v${VERSION}"
+
+          # 2. Recreate the release branch from the latest main.
```
rchiodo: 📍 .github/workflows/shared-package-submodule-sync.yml:62
The explicit lease is a real improvement over the prior bare form, but the carried-forward data-loss path persists (prior High, still unresolved). The Skeptic re-reproduced it against a bare remote: on a same-tag re-dispatch after a maintainer pushes a fixup onto `shared-package-vX.Y.Z`, `PRIOR_REMOTE` captures that fixup, so the `--force-with-lease="${BRANCH}:${PRIOR_REMOTE}"` push trivially passes and `checkout -B "${BRANCH}" "${MAIN_SHA}"` + force-push silently overwrites it (push exited `0`, `maintainer commit LOST`). The lease guards the in-run race window, not pre-existing branch commits, and the `git diff --quiet` short-circuit doesn't help because the rebuild-from-main always shows a submodule diff. The comment ("so a concurrent update is not silently overwritten") overstates the protection. Fix: commit onto the existing branch (`git checkout -B "${BRANCH}" "origin/${BRANCH}"` when `PRIOR_REMOTE` is non-empty) or short-circuit when the branch already has an open tracking issue; if force-regenerate-from-main is the deliberate policy, drop the lease and comment and use an honest `--force`.

[verified]

noxfile.py
```diff
@@ -110,6 +110,16 @@ def install_bundled_libs(session):
...
     _install_bundle(session)
+    # Source the shared Python library from the git submodule instead of the
+    # published package so the bundled copy matches the pinned submodule commit.
+    session.install(
+        "-t",
+        "./bundled/libs",
+        "--no-cache-dir",
+        "--no-deps",
+        "--upgrade",
+        "./external/vscode-common-python-lsp/python",
```
rchiodo, line 127: 📍 noxfile.py:113
Installing the vendored shared Python package with `--no-deps` while its hash-pinned entry was removed from `requirements.txt` means any runtime dependency of `./external/vscode-common-python-lsp/python` not already in `requirements.in` (`pygls`, `packaging`, `flake8`) will not land in `bundled/libs`, producing a runtime `ImportError` invisible at build time. Drop `--no-deps` so pip resolves the package's declared deps, or enumerate them back into `requirements.in` so they stay hash-pinned and bundled.

[verified]

rchiodo:
Main themes: (1) the release-branch force-push can overwrite maintainer commits despite the lease, and (2) `--no-deps` may drop runtime deps of the vendored Python package from the bundle. Also please update the PR description to match the actual submodule-sync workflow.

noxfile.py
rchiodo, line 127: The vendored shared package is installed with `--no-deps`, and its hash-pinned entry was removed from `requirements.txt`. `pygls`/`packaging`/`lsprotocol` are now pinned in `requirements.in` to compensate, but this is a manual contract: any future runtime dependency added upstream to `external/vscode-common-python-lsp/python` will silently miss `bundled/libs` and surface as a runtime `ImportError` with no build-time signal. Consider dropping `--no-deps` so pip resolves declared deps, or adding a build-time check that the submodule's declared runtime deps are a subset of `requirements.in`.

rchiodo:
Looks good overall. Two non-blocking notes worth addressing: the PR description doesn't match the actual implementation (git-submodule vendoring, not an npm/uv recompile workflow), and the `--no-deps` install of the vendored Python lib relies on a manual dependency contract worth guarding.

rchiodo (dismissed):
Approved via Review Center.

bschnurr (dismissed):

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,203 @@
...
+          echo "BASH_ENV=${HELPER}" >>"${GITHUB_ENV}"
+
+      # Once the release branch exists a maintainer may be editing it, so the
+      # remaining steps are skipped to avoid clobbering their work.
+      - name: Check whether the release branch already exists
+        id: guard
+        run: |
+          set -euo pipefail
+          if git ls-remote --exit-code --heads origin "refs/heads/${BRANCH}" >/dev/null 2>&1; then
+            echo "Branch ${BRANCH} already exists; leaving it untouched."
```
rchiodo, line 120: 📍 .github/workflows/shared-package-submodule-sync.yml:90
Re-dispatch can permanently drop the maintainer notification. If run #1 pushes `${BRANCH}` (step 6) but `gh issue create` then fails, it is swallowed by `|| echo` and the job exits green with no tracking issue. Any re-dispatch of the same tag hits this `git ls-remote` branch-exists guard and `exit 0`s before ever reaching the issue block, so the branch sits forever with no PR and no notification. Fail-safe (no data loss), but recovery requires manually deleting the branch. On the branch-exists path, still reconcile/ensure an open tracking issue before `exit 0` (or make issue-creation failure non-swallowed so the run is red and retriable).

rchiodo:
Solid submodule-sync migration. Two non-blocking robustness notes worth addressing before relying on the automation in anger; otherwise good to merge.

rchiodo (dismissed):
Approved via Review Center.

rchiodo:
CI hardening and the submodule-vendoring approach look solid. The dropped candidates were mostly duplicates or trusted-payload DoS-only nitpicks. Two worthwhile non-blocking follow-ups and a PR-description fix are noted inline.

rchiodo (dismissed):
Approved via Review Center.

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,247 @@
...
+          fi
+
+      # The extension builds the shared package's TypeScript during install, so its
+      # Node toolchain must satisfy the package's declared build floor
+      # (engines.node), not the exact dev pin in the submodule .nvmrc. An
+      # unparseable floor only warns rather than failing every future sync.
+      - name: Verify Node version compatibility
+        if: steps.guard.outputs.exists == 'false' && steps.detect.outputs.changed == 'true'
+        run: |
+          set -euo pipefail
```
rchiodo, line 212: 📍 .github/workflows/shared-package-submodule-sync.yml:187
Unresolved from prior review (Skeptic + Architect verified). The issue step is gated `exists=='false' && changed=='true'`, and `gh issue create` failure is swallowed by `|| echo`. If run #1 pushes the branch but the issue create fails, any re-dispatch of the same tag hits the branch-exists guard and `exit 0`s before the issue block — the branch sits forever with no PR and no notification (recovery requires manually deleting the branch). Fail-safe (no data loss), but the maintainer notification is permanently lost. On the branch-exists path, reconcile/ensure an open tracking issue before exiting, or make issue-creation failure non-swallowed so the run is red and retriable.

[verified]

.github/workflows/pr-check.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
```
rchiodo, line 21: 📍 .github/workflows/pr-check.yml:20
Architect (Medium, maintenance risk not a bug): the recursive-submodule-init invariant is now duplicated across ~13–14 sites (4 checkouts here, 4 in push-check.yml, 5 Azure pipeline files, plus the sync workflow). Because `package.json` uses a `file:` link, any future workflow/pipeline that omits `submodules: recursive` silently breaks at `npm ci`/webpack. Every current site is correct, but consider consolidating into a single composite/local checkout action so the invariant can't be forgotten later.

[verified]

rchiodo:
Solid submodule-vendoring setup. Non-blocking: the sync workflow can lose the maintainer notification if branch-push succeeds but issue-creation fails, and the `--no-deps` install turns `requirements.in` into a manual mirror of the submodule's runtime deps. Also update the PR description to match the actual workflow file.

rchiodo (dismissed):
Approved via Review Center.

.github/workflows/shared-package-submodule-sync.yml
```diff
@@ -0,0 +1,247 @@
...
+            echo "::error::Extension Node version (${EXT_NODE}, from .nvmrc) is older than the shared package's required build floor (>=${PKG_NODE}, from engines.node). Update .nvmrc, or investigate the shared release before syncing."
+            exit 1
+          else
+            echo "Node versions compatible: extension ${EXT_NODE} >= shared package build floor ${PKG_NODE}."
+          fi
+
+      # Refresh the lockfile so the branch stays `npm ci`-mergeable, then commit
+      # both the submodule pointer and the regenerated lockfile. The branch is
+      # new (guarded above), so a plain push cannot overwrite anyone's work.
+      - name: Commit and push the update
```
rchiodo, line 230: The Node-compatibility step's documented "warn and skip on unparseable floor" fallback is unreachable. `EXT_NODE="$(grep -Eo … .nvmrc | head -n1)"` and `PKG_NODE="$(printf … | grep -Eo …)"` run under `set -euo pipefail`: `grep` exits 1 on no match, and with `pipefail` the pipeline inherits that failure, aborting the step in a top-level assignment *before* the `if [ -z "${PKG_NODE}" ]` warn-branch. So a future shared release with a missing/non-numeric `engines.node` (or a `.nvmrc` alias like `lts/*`) wedges the sync with an opaque red run — the opposite of the comment's promise. The Python step is immune because it uses `sed -n '…p'` (exits 0 on no match). Fix: append `|| true` to the `EXT_NODE`/`PKG_NODE` assignments (as already done for `PKG_RANGE`), or use `sed -n`, so the `-z` warn-branch behaves as documented.

package.json
```diff
@@ -54,6 +54,7 @@
     "main": "./dist/extension.js",
     "l10n": "./l10n",
     "scripts": {
+        "postinstall": "node ./build/postinstall.js",
```
rchiodo, line 57: Maintenance-risk note (not a bug): the `file:` link plus the `postinstall` build couples the extension's install to the shared package's *entire* dev toolchain — `package-lock.json` now hoists `typescript@^6.0.3` (a major preview), `mocha@11`, `sinon@22`, etc. into the extension's lockfile, so upstream devchain churn can break the extension's `npm ci`. Consider shipping prebuilt submodule output, or pinning/isolating the submodule's build toolchain, to decouple the extension build from the shared package's evolving devchain.

rchiodo:
Overall this is a solid, well-guarded change. A couple of edge-case reliability notes on the sync workflow (unreachable warn-branch in the Node compat check, and a permanently-dropped tracking-issue notification on the branch-exists path) are worth addressing, plus two architectural notes about the `file:` link coupling and the manual `--no-deps` contract. None are blocking.

rchiodo (dismissed):
Approved via Review Center.

rchiodo:
🔒 Automated review in progress — @rchiodo is auto-reviewing this PR.

edvilme:
## Review fixes applied (6 commits)

Pushed 6 identical commits across all 5 consumer PRs (one per fix). The workflow/`postinstall.js`/`dependabot.yml` changes are byte-identical in every repo:

1. **`sync.yml` tag resolution** — added `refs/tags/${VERSION}` (bare) as a third candidate so `v`-prefixed and bare release tags both resolve symmetrically.
2. **`sync.yml` `requires-python` floor** — rewrote the parse to extract the quoted value and read only the `>=`/`~=` lower bound (empty → warn-skip) instead of greedily grabbing the first `major.minor`.
3. **`sync.yml` ref hardening** — added `git check-ref-format "refs/heads/${BRANCH}"` to reject multiline / `..` injection before emitting outputs.
4. **`sync.yml` tracking-issue step** — quoted the `in:title` search and stopped swallowing `gh` failures (removed `2>/dev/null || true`) so errors surface under `set -e`.
5. **`postinstall.js`** — skip the build when `dist/index.js` already exists, and skip-with-guidance (instead of hard-failing) when the submodule TS toolchain is absent (dev-pruned installs).
6. **`dependabot.yml`** — annotated the two now-inert shared-package ignore entries as intentional defensive no-ops.

### Deferred (need human/CI judgement — left unresolved intentionally)
Design/verification-level threads are **not** auto-fixed: lockfile drift from `npm install --package-lock-only`, the compat-guard "validates the wrong minimum" redesign (runtime.txt/.nvmrc vs constants), the noxfile `--no-deps` cross-repo coupling guard, project-wide `tsconfig skipLibCheck`, the `file:` vendoring/`.vscodeignore` bundling verification, and the sync-notification/branch-exists robustness gap. These warrant maintainer review rather than a mechanical change.

edvilme:
## Follow-up review fixes (3 more commits)

Three additional identical commits address further review feedback:

7. **`sync.yml` — re-notify on re-dispatch.** The "Open or update the tracking issue" step now also runs when the sync branch already exists (`exists == 'true'`). If a prior run pushed the branch but died before notifying (or the tracking issue was closed), a re-dispatch now re-ensures the tracking issue and re-emits the compare URL to the job summary instead of silently `exit 0`-ing. The step is idempotent (it reuses any open issue). Combined with the earlier removal of the swallowed `gh` failure, the notification can no longer be permanently lost.
8. **`sync.yml` — resolve symbolic Python floor.** The Python-compat gate now falls back to resolving `MINIMUM_PYTHON_MAJOR`/`MINIMUM_PYTHON_MINOR` when `minimumPythonVersion` is declared via named constants (as in vscode-pylint), so the gate is no longer a silent no-op there.
9. **`noxfile.py` — submodule guard.** Added a friendly missing-submodule check (`git submodule update --init --recursive`) before installing the shared Python lib, symmetric with the `build/postinstall.js` guard.

### Already addressed in the current workflow (earlier-round threads resolved)
Several earlier comments referenced code that in-place hardening already fixed: the `ls-remote` branch-exists check now retries and treats only exit code 2 as "absent"; the Node floor is read from `engines.node` (not `.nvmrc`); the Python floor is compared against `src/common/constants.ts` (not `runtime.txt`); and an unparseable floor now warns-and-skips while the compat steps are gated on an actual submodule change.

### Still open for maintainer decision (not auto-changed)
Lockfile drift from `npm install --package-lock-only`, whether the sync should build/validate (`npm ci` + `nox install_bundled_libs`) before declaring the branch "ready" vs. wording the issue as "needs validation", the noxfile `--no-deps` runtime-dep mirroring strategy, preventing `nox -s update_packages` from rewriting the `file:` dependency, project-wide `tsconfig skipLibCheck`, and the `.vscodeignore`/`pr-check` verification items. These are design/verification calls I've left for review.

.github/workflows/pr-check.yml
edvilme: Acknowledged as a maintenance-risk note rather than a bug. The recursive-submodule-init requirement is asserted in both pr-check and the sync workflow because they run in independent contexts; keeping both avoids a silent regression. Resolving.

package.json
edvilme: Acknowledged (not a bug). Coupling the extension install to the shared package build via `file:` + postinstall is the intended trade-off of submodule consumption; the packaging pipeline runs a full build. Resolving.

noxfile.py
edvilme: Acknowledged as an intentional design choice. `--no-deps` is required because the shared package installs from a local path and `--generate-hashes` cannot hash a local path; its runtime deps (pygls/lsprotocol/packaging) stay pinned and hashed in the consumer's own requirements. Keeping them in sync is a deliberate, reviewed step of a shared-package release rather than something we auto-mirror. Resolving.

edvilme:
## Review follow-up complete

All review threads on this PR are now resolved. Latest round of identical fixes across all five consumer PRs:

- **Floor-parse hardening** — Node/Python compatibility gates now warn-and-skip an unparseable floor instead of aborting the step under `set -euo pipefail`.
- **`nox -s update_packages` safety** — `@vscode/common-python-lsp` is pinned so the updater can't rewrite the `file:` link back to a registry range.
- **Honest postinstall comment** — clarified that a non-recursive clone fails at `file:` resolution during install, not at postinstall.
- **Deterministic tracking issue** — the tracking-issue title now uses the canonical `v${VERSION}`, so re-dispatching the same release as `1.2.3` and `v1.2.3` reuses one issue instead of duplicating it; the issue body now states the pushed branch is unvalidated and must pass PR checks before merge.

Remaining review notes (lockfile `--package-lock-only` re-resolution, Python `--no-deps` dependency mirroring, project-wide `skipLibCheck`, and the dispatch-only compatibility gate) were acknowledged inline as documented, intentional trade-offs of the submodule consumption model.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [noxfile.py:L628](https://github.com/microsoft/vscode-flake8/blob/c1225f01bc364eafd1f4f031a43baa88b5e7e762/noxfile.py#L628).

Installing the shared Python package with `--no-deps` means its runtime dependencies must stay manually mirrored in `requirements.in`. A future automatically synced release that adds a dependency can bundle successfully but fail with `ImportError` at runtime, and the sync workflow does not validate that dependency set. Please add a check that the shared package's runtime dependencies are covered by the pinned requirements, or otherwise ensure they are installed from the submodule metadata.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

